Issue #10, opened 2023-11-14
Weird deformation of cylinder caps and non-correct render of planes (correlated?)


u413-284-si:
If the cylinder is rendered off-image-centre, its end caps are deformed and do not conform expectations.

As the caps are checked with exactly the same code for intersection as the planes (which also are not rendered correctly) an assumption is made that the root problems lies at this point.

Further investigation required, potentially facilitated by the added functionality of being able to move objects in space via keys. Therefore being able to analyse different viewing angles and positions quickly instead of hard coding values, then compiling.

Issue #7, opened 2023-10-23
World creation


u413-284-si:
The scene to be interacted with and displayed needs to be modeled. This shall be performed after the parser has obtained all information required. 

Contains:

- Translate all data from *.rt file into model world
- Add optional elements into world
- World may be an overarching struct containing substructs defining each element

u413-284-si:
Done with parser-branch

Issue #3, opened 2023-10-23
Parser


u413-284-si:
First argument after programm call is input file to read (format: *.rt). 

Contains:

> following elements
- Ambient lighting: A   ratio[0.0 - 1.0]   R G B[0 - 255]
- Camera: C   view_point_coordinates   normalised_orientation_vector[-1, 1; x, y, z]   field_of_view[0, 180°]
- Light: L   light_point_coordinates   brightness_ratio[0.0 - 1.0]   (R G B[0 - 255])
- Sphere: sp   sphere_centre_coordinates   sphere_diameter   R G B[0 - 255]
- Plane: pl   point_in_plane_coordinates   normalised_normal_vector[-1, 1; x, y, z]   R G B[0 - 255]
- cylinder: cy    cylinder_centre_coordinates   normalised_axis_vector[-1, 1; x, y, z]   cylinder_diameter   cylinder_height   R G B[0 - 255]

Conditions:

- elements may be separated by  >= 1 line breaks
- token in elements may be separated by >= 1 spaces
- element order arbitrary
- element declaration by capital letter are unique (no further declarations for the scene)
- element headed by identifier (1 or 2 letters) followed by element information in strict order

Target parser behaviour:

- save element information inside data struct(s) with appropriate variables
- struct readily available to raytracer
- misconfigurations of *.rt file lead to "Error\n" output + error message of choice

u413-284-si:
When reading light colours from file the RGB values need to be converted into the range [0, 1] (therefore read value / 255). The definitionf of t_colour is as follows:

/**
 * @brief Represents an RGB colour with float values from 0.0 - 1.0
 * @param r		Red component
 * @param g		Green component
 * @param b		Blue component
 */
typedef struct s_colour
{
	float	r;
	float	g;
	float	b;
}	t_colour;

And the light colour shall be saved as 

/**
 * @brief Represents lighting in the scene
 * @param pos		Position of the light
 * @param ratio		Light brightess ratio
 * @param colour	Colour of the light
 */
typedef struct s_light
{
	t_vec3		pos;
	float		ratio;
	t_colour	colour;
}	t_light;

PR #11, opened 2023-11-14
5 ray tracer

### LINKED ISSUE #5: Ray tracer
The ray tracer is responsible to define the behaviour of rays inside the scene/world by describing their interactions with the different elements pertaining to it. The rays eventually return information on the elements by defining a ray colour (RGB).

In order to obtain enough (colour) information for each pixel of the viewport (virtual rectangle containing the grid of equidistantly spaced image pixel locations - horizontally and vertically) at least one ray is sent per pixel into the scene. For higher resolutions and anti-aliasing more ray samples may be send per pixel. 

Contains:

- ray creation and collection ? (see camera)
- ray colourisation
- intersection behaviour/hit record for
   * sphere
   * plane
   * cylinder
- lighting:
   * hard shadows/determine whether object in shadow
   * ambient lighting
   * diffuse lighting

u413-284-si:
Ray tracer implemented with features as specified in Issue #5. Therefore closes #5 .

Includes:
* General struct hittables containing all objects implemented to date: spheres, planes, cylinders -> can be extended with further shapes
* Lighting: ambient, diffuse, shadowing (hard)
* Colour attribution depending on lighting and object colour
* Hit calculation based on solving a quadratic equation
* Rudimentary scene creation to be replaced with Issues #3 and #7

Open points: 
* Issue #10 
* Code is yet to be formatted for 42


.vscode/launch.json
```diff
@@ -0,0 +1,33 @@
...
+	"version": "0.2.0",
+	"configurations": [
+		{
+			"name": "Standard",
+			"type": "cppdbg",
+			"request": "launch",
+			"program": "${workspaceFolder}/miniRT",
+			"args": ["er", ">", "image.ppm"],
+			"stopAtEntry": true,
+			"cwd": "${fileDirname}",
```
gwolf-011235: change to "workspaceFolder" 
- works even if no file selected
-  output gets saved at project root directory

.vscode/tasks.json
```diff
@@ -0,0 +1,33 @@
...
+			],
+			"options": {
+				"cwd": "${workspaceFolder}"
+			},
+			"problemMatcher": [
+				"$gcc"
+			],
+			"group": "build",
+		}
+	]
```
gwolf-011235: Can be deleted, is alternative config for "make test" - there is no test target

inc/camera.h
gwolf-011235, line 1: General: t_cam is already very big. As far as I can tell are many variables only used once in a function. I suspect some variables are set for "future use" - but then I'd suggest to add them when they are needed. It makes it rather confusing why something is there, and if its intended for later use or was an oversight. Also general: do we really need to save them all for later use? Example:
- camera_centre is set to look_from. Look_from is the position the camera is looking from. Camera_centre has no doxy. I think they are the same?
- viewport_w is never used (I think because we need it later). It is calculated with w, which is also saved in the struct. Do we need to store w? Couldn't it be just a function variable?
- vfov stores a degree value. In ft_initiate_viewport() a temp var converts it to radian. 

I also think that t_cam should be split off from viewport, as you already do with ft_initiate_camera() and ft_initiate_viewport(). Could help with readability.

inc/hit_manager.h
```diff
@@ -18,10 +18,28 @@
...
+ * occured or not
+ * 
+ * @param obj		Hittable in question
+ * @param ray 		Ray shot into scene
+ * @param rec		Hit record of ray with hittable
+ * @param ray_d 	Interval of accepted values of the ray distance for a hit
+ * @return true		If hittable is hit
+ * @return false	If hittable is missed
+ */
+bool	ft_hit_scene(t_hittable obj, t_ray ray, t_hitrecord *rec, \
```
gwolf-011235: ft_hit_object / ft_hit_shape / ft_hit_hittable
It only hits one thing, not entire scene.

inc/utils.h
```diff
@@ -65,4 +89,29 @@ bool	ft_contains(float x, t_interval interval);
...
 bool	ft_surrounds(float x, t_interval interval);
 
+/**
+ * @brief Returns the bigger of two values
+ * 
+ * @param x 	First value
+ * @param y 	Second value
+ * @return float 
+ */
+float	ft_max(float x, float y);
```
gwolf-011235: Why not fmaxf() from <math.h> ? Performance?

inc/scene.h
```diff
@@ -25,6 +25,36 @@
...
 
+/**
+ * @brief Shows point of intersection
+ * @param point		Point of intersection
+ * @param normal	Normal vector at point of intersection
+ * @param axis_hit	Point of intersection with a cylinders axis
+ * @param d			Distance into ray direction when point is hit
+ * @param colour	Colour of object at intersection
+ */
+typedef struct s_hitrecord
```
gwolf-011235: I think this was a better fit in ray.h - it records the hit of a ray with an object.

inc/scene.h
```diff
@@ -25,6 +25,36 @@
...
+{
+	t_vec3		point;
+	t_vec3		normal;
+	t_vec3		axis_hit;
```
gwolf-011235: Is an axis_hit (which is only needed for cylinder?) different than a normal hit?

inc/scene.h
```diff
@@ -63,28 +95,77 @@ typedef struct s_cylinder
...
+}	t_shape;
+
+/**
+ * @brief Represents a hittable 
+ * 
+ * @param id		The hittables ID
+ * @param type		The hittables type
+ * @param params	Required parameters to distinctly describe the hittable
+ */
+typedef struct s_hittable
```
gwolf-011235, line 125: Why not object? In all functions it's called obj.
Or shape and renme t_shape into t_object.

inc/scene.h
```diff
@@ -63,28 +95,77 @@ typedef struct s_cylinder
...
 	float		h;
 	t_colour	colour;
 }	t_cylinder;
 
+/**
+ * @brief Contains numerical list of possible hittables
+ * 
+ * Starts at 0.
+ */
+typedef enum e_type
```
gwolf-011235, line 93: I think I read that to guarantee that list starts with zero, at least one (i.e. the first) should be explicitly assigned: "SPHERE = 0,"

Also: in check module I'm using a similiar enum (ent_type), which has all entity types (+ambient, camera, light). Could we fuse them?

inc/scene.h
gwolf-011235, line 1: I suggest that this header only contains the entities. Which means moving the functions out into a init.h. This would make scene.h more standalone. Also renaming in entities.h would make it very clear whats defined in there

inc/utils.h
gwolf-011235, line 1: More like math_utils :O

scenes/01_basic_shapes.rt
gwolf-011235, line 1: Did you use this? Wouldn't you have to comment in and out stuff? I would make 4 files out of it.

src/colour.c
```diff
@@ -23,3 +23,21 @@ void	ft_write_colour(t_colour pixel)
 	ib = (int)(255.99 * pixel.b);
 	printf("%d %d %d\n", ir, ig, ib);
 }
+
+t_colour	ft_add_colour(t_colour c1, t_colour c2)
+{
+	return ((t_colour){
+		.r = c1.r + c2.r,
+		.b = c1.b + c2.b,
+		.g = c1.g + c2.g,
```
gwolf-011235: I think nominette accepts this, which I think is a bug -> norminette doesn't check return(). 
I'd use a variable. Since you already pass by value you could also c1.r += c2.r

src/colour.c
```diff
@@ -23,3 +23,21 @@ void	ft_write_colour(t_colour pixel)
...
+	});
+}
+
+t_colour	ft_hadamard_colour(t_colour c1, t_colour c2)
+{
+	return ((t_colour){
+		.r = c1.r * c2.r,
+		.b = c1.b * c2.b,
+		.g = c1.g * c2.g,
```
gwolf-011235: Same as above

src/ray.c
```diff
@@ -20,26 +20,38 @@ t_vec3	ft_ray(t_ray ray, float d)
...
+		ft_enlighten(&ray_colour, rec, scene);
+		return (ray_colour);
+	}
+/* 	{
 		sphere_colour.r = 0.5 * (rec.normal.x + 1.0);
 		sphere_colour.g = 0.5 * (rec.normal.y + 1.0);
 		sphere_colour.b = 0.5 * (rec.normal.z + 1.0);
 		return (sphere_colour);
-	}
+	} */
```
gwolf-011235: Can be deleted I think

src/ray.c
```diff
@@ -20,26 +20,38 @@ t_vec3	ft_ray(t_ray ray, float d)
...
 
 	// Background
```
gwolf-011235: This background stuff could be moved out into it's own function. How we color the background (solid or maybe change depending on some keypress) can be later adapted

src/scene.c
gwolf-011235, line 1: I think init.c or test_init.c would be better

src/scene_light.c
```diff
@@ -0,0 +1,58 @@
...
+	}
+}
+
+t_colour	ft_ambient_light(t_light ambient)
+{
+	return ((t_colour){
+		.r = ambient.colour.r * ambient.ratio,
+		.b = ambient.colour.b * ambient.ratio,
+		.g = ambient.colour.g * ambient.ratio,
+	});
```
gwolf-011235, line 43: Like above: assignment on return

src/scene_light.c
```diff
@@ -0,0 +1,58 @@
...
+	t_vec3	light_dir;
+	float	dot_product;
+
+	light_dir = ft_vec3_norm(ft_vec3_sub(cur.pos, rec.point));
+	dot_product = ft_max(0.0, ft_vec3_dot(light_dir, rec.normal));
+	return ((t_colour){
+		.r = dot_product * cur.ratio * cur.colour.r,
+		.b = dot_product * cur.ratio * cur.colour.b,
+		.g = dot_product * cur.ratio * cur.colour.g,
+	});
```
gwolf-011235, line 57: Like above: assignment on return

src/scene_shadow.c
```diff
@@ -0,0 +1,44 @@
...
+
+#include "lighting.h"
+
+bool	ft_in_shadow(t_light cur, t_hitrecord rec, t_entities scene)
+{
+	t_ray		shadow_ray;
+	t_vec3		light_dir;
+	t_interval	ray_d;
+
+	light_dir = ft_vec3_sub(cur.pos, rec.point);
```
gwolf-011235, line 21: Should be normalized (since it's direction)?

inc/hit_manager.h
gwolf-011235, line 1: I would suggest to split every hit function of the different shapes into it's own header. 
Alternative: add an identifier (eg: ft_sp_ for sphere and ft_cy for cylinder) to make it easy to distinguish: ft_check_wall() is a cylinder only function.
Or both.

src/hit_cylinder.c
```diff
@@ -0,0 +1,97 @@
...
+	return (true);
+}
+
+bool	ft_calculate_pot_hits(t_cylinder cy, t_ray ray, t_interval ray_d, \
+	float potential_hits[4])
+{
+	t_equation	eq;
+	t_vec3		cap1_ray;
+	float		d3;
+	float		d4;
```
gwolf-011235: nit: cap_hit_top and cap_hit_bot instead of d3 and d4?

src/utils_cylinder.c
```diff
@@ -0,0 +1,77 @@
...
+}
+
+bool	ft_visible(t_equation eq, t_interval ray_d, float d3, \
+	float d4)
+{
+	if (!ft_surrounds(eq.d1, ray_d))
+		if (!ft_surrounds(eq.d2, ray_d))
+			if (!ft_surrounds(d3, ray_d))
+				if (!ft_surrounds(d4, ray_d))
+					return (false);
```
gwolf-011235, line 75: Why not just one If-Statement with && ?

src/utils_cylinder.c
gwolf-011235, line 1: I think these are also hit functions for cylinder, I'd suggest rename to hit_cylinder2.c

src/vec3_arithmetics.c
```diff
@@ -6,40 +6,37 @@
...
-	t_vec3	rtrn;
-
-	rtrn.x = v1.x + v2.x;
-	rtrn.y = v1.y + v2.y;
-	rtrn.z = v1.z + v2.z;
-	return (rtrn);
+	return ((t_vec3){
+		.x = v1.x + v2.x,
+		.y = v1.y + v2.y,
+		.z = v1.z + v2.z,
```
gwolf-011235, line 20: Change as mentioned above

src/vec3_arithmetics.c
```diff
@@ -6,40 +6,37 @@
...
-
-	rtrn.x = v1.x - v2.x;
-	rtrn.y = v1.y - v2.y;
-	rtrn.z = v1.z - v2.z;
-	return (rtrn);
+	return ((t_vec3){
+		.x = v1.x - v2.x,
+		.y = v1.y - v2.y,
+		.z = v1.z - v2.z,
+	});
```
gwolf-011235, line 30: Change as mentioned above

src/vec3_arithmetics.c
```diff
@@ -6,40 +6,37 @@
...
-	t_vec3	rtrn;
-
-	rtrn.x = v.x * a;
-	rtrn.y = v.y * a;
-	rtrn.z = v.z * a;
-	return (rtrn);
+	return ((t_vec3){
+		.x = v.x * a,
+		.y = v.y * a,
+		.z = v.z * a,
```
gwolf-011235, line 38: Change as mentioned above

src/vec3_arithmetics.c
```diff
@@ -49,10 +46,9 @@ float	ft_vec3_dot(t_vec3 v1, t_vec3 v2)
...
-
-	rtrn.x = v1.y * v2.z - v1.z * v2.y;
-	rtrn.y = v1.z * v2.x - v1.x * v2.z;
-	rtrn.z = v1.x * v2.y - v1.y * v2.x;
-	return (rtrn);
+	return ((t_vec3){
+		.x = v1.y * v2.z - v1.z * v2.y,
+		.y = v1.z * v2.x - v1.x * v2.z,
+		.z = v1.x * v2.y - v1.y * v2.x,
+	});
```
gwolf-011235, line 53: Change as mentioned above

src/vec3_linalgebra.c
```diff
@@ -29,3 +29,17 @@ float	ft_vec3_angle(t_vec3 v1, t_vec3 v2)
...
+t_vec3	ft_vec3_hadamard(t_vec3 v1, t_vec3 v2)
+{
+	return ((t_vec3){
+		.x = v1.x * v2.x,
+		.y = v1.y * v2.y,
+		.z = v1.z * v2.z,
+	});
+}
+
+bool	ft_vec3_equal(t_vec3 v1, t_vec3 v2)
```
gwolf-011235, line 42: This function is extremely spicy: equal comparison of floats is almost never want you want to do

src/vec3_linalgebra.c
```diff
@@ -29,3 +29,17 @@ float	ft_vec3_angle(t_vec3 v1, t_vec3 v2)
 	dot_product = ft_vec3_dot(v1, v2);
 	return (acos(dot_product / (ft_vec3_abs(v1) * ft_vec3_abs(v2))));
 }
+
+t_vec3	ft_vec3_hadamard(t_vec3 v1, t_vec3 v2)
+{
+	return ((t_vec3){
+		.x = v1.x * v2.x,
+		.y = v1.y * v2.y,
+		.z = v1.z * v2.z,
```
gwolf-011235, line 38: Change as mentioned above

gwolf-011235 (changes requested):
Projection of different objects looks nice!
The abstract t_hittable array is a good choice for sure.
Two functions check equality of floats, I think this has to be addressed.
Rest of the comments are more or less nit: naming convention and stuff

.vscode/launch.json
u413-284-si: mu

.vscode/tasks.json
u413-284-si: mumu

inc/camera.h
u413-284-si: t_cam split into t_image, t_cam and t_viewport

camera_centre and lookfrom seem to be the same. Lookfrom is kept until further insight is gained.

viewport_w is indeed unused->deleted

vfov changed to hfov. will be converted to radian while parsing

inc/hit_manager.h
u413-284-si: changed to ft_hit_hittable()

inc/utils.h
u413-284-si: wasn't aware of it, replaced now

inc/scene.h
u413-284-si: k

inc/scene.h
u413-284-si: yes, that's basically the projection of the hit point unto the axis

inc/scene.h
u413-284-si: Cause an object would include lights, but these are not hittable

inc/scene.h
u413-284-si: yeah, make it one when we merge your branch

inc/scene.h
u413-284-si: done

u413-284-si: reasonable suggestion

inc/utils.h
u413-284-si: naaaa, is gut name

scenes/01_basic_shapes.rt
u413-284-si: not yet, wanted to have it as ref for testing

src/ray.c
u413-284-si: is deleted

src/ray.c
u413-284-si: is in function

src/scene.c
u413-284-si: renamed it scene_init.c

src/scene_shadow.c
u413-284-si: it is normalised three lines after, but need the extra variable to calculate its length

inc/hit_manager.h
u413-284-si: added identifier to all cylinder functions a la ft_cy_....

src/hit_cylinder.c
u413-284-si: we dont know which is top and which bottom as the cylinder orientation can be anything. Also the first two solutions are named d1 and d2, therefore I'd keep it consistent

src/utils_cylinder.c
u413-284-si: Cause I want to check each solution sequentially, and if one is not fullfilled, not check the underlying ones

src/utils_cylinder.c
u413-284-si: But that's ugly, just giving numbers to file names. As I classify these functions as helper functions, I put them under that name

src/vec3_linalgebra.c
u413-284-si: True, any suggestions? Somehow compare only to a certain decimal place?

u413-284-si: Implemented a solution with a threshold

u413-284-si:
Most change request implemented. Awaiting decision on usage of compound literals.

inc/utils.h
gwolf-011235: I heard that the point of math.h is precision, which is rather slow. Since we don't really care about that, we could also use a self made version.
I don't know if it really matters with max() though

inc/camera.h
gwolf-011235: Made an issue to not forget conversion for parsing. 

Is calc changed to use horizontal fov?

src/utils_cylinder.c
gwolf-011235: If-conditions are guaranteed by c standard to be checked from left to right. 

Consider the following:
if (cond1 && cond2) {};
cond2 will never be checked if cond1 is false.

This means it's possible to do the following:
if (ptr != NULL && ptr->ref) {};
Even if ptr is NULL it's guaranteed, that's it's not dereferenced, since it's the second condition.

inc/camera.h
```diff
@@ -16,75 +16,117 @@
...
+typedef struct s_viewport
 {
-	float	aspect_ratio;
-	int		image_width;
-	int		image_height;
 	float	viewport_width;
 	float	viewport_height;
 	float	focal_length;
-	t_vec3	camera_centre;
+	float	hfov;
```
gwolf-011235: hfov should be in t_cam. It is used in viewport init, but gets parsed into camera

src/hit_cylinder.c
gwolf-011235: I thought the "d" of d1 and d2 stand for discriminant. Turns out its just the next letter after c :D
But still: d3 and d4 are not solutions of a quadratic formula. Therefore I think it's precisely not consistent to name them that way.
cap_hit1 and cap_hit2 ignores orientation.

But again: just a nit, we can leave it.

inc/camera.h
```diff
@@ -16,75 +16,117 @@
...
- * @brief Contains camera, image and viewport parameters
+ * @brief Contains image parameters
+ * 
+ * @param aspect_ratio			Ratio of image_width / image_height
+ * @param image_width			Width of image created
+ * @param image_height			Height of image created
+ */
+typedef struct s_image
+{
+	float	aspect_ratio;
```
gwolf-011235, line 32: Since we define height and width ourselves (hardcoded, screen width, or maybe as optional arg), we don't need aspect ratio to calc image size.
But we could maybe have some kind of transformation, which changes the aspect ratio. 
Should be keep it in for maybe later use? If yes, I will create issue to not forget

src/utils_cylinder.c
gwolf-011235: k

src/vec3_linalgebra.c
gwolf-011235: I think it should be abs of the difference i.e. fabs(v1.x - v2.x)

At least according to the "simple solution" of this thread by [nni6] https://stackoverflow.com/questions/4915462/how-should-i-do-floating-point-comparison/19050413#19050413
(and some others online also take abs()).

In the same thread one commenter says no fabs() needed if negative Epsilon is used: A - B < -Epsilon 

The two top answers seem to be over the top for our usecase.

Wrapper suggestion:

bool ft_nearly_equal_flt(float one, float two) {
if (fabs(one -two) < EPSILON)
return (true);
return (false); }

src/utils.c
```diff
@@ -26,3 +26,20 @@ bool	ft_surrounds(float x, t_interval interval)
 {
 	return (interval.min < x && interval.max > x);
 }
+
+float	ft_solve(t_equation *eq)
+{
+	float	discriminant;
+
+	if (eq->a == 0.0 && eq->b != 0.0)
```
gwolf-011235, line 34: Spicy equality comparison of floats? Should it also return if nearly zero or only if zero? If later then ok.

src/colour.c
gwolf-011235: Let's leave it. I will create a new issue, so we don't forget if they respond

inc/camera.h
u413-284-si: yes, the code has already been changed to calculate hfov

inc/utils.h
u413-284-si: I would leave it as is. Don't think it will have any effect on our little prog

inc/camera.h
gwolf-011235: top

inc/utils.h
gwolf-011235: k

src/utils_cylinder.c
u413-284-si: yeah, that might be, but I think the readability/message of the sequential verification is clearer with the current version. The other one requires pre-knowledge about this behaviour and one could easily misunderstand that all conditions are checked "simultaneously". Therefore I'd leave it as is

gwolf-011235: k

gwolf-011235 (commented):
Elegant split of t_cam. 
I think hfov should be moved into t_cam.
Solved comments about compound literal and made issue.
One more equality check for float found.
If equality check for float is good can be merged.

src/hit_cylinder.c
u413-284-si: I guess we could argue that till the end of days as each perspective has its merit. To put an end to it and as a my purpose in life is to make you happy, I renamed the two additional solutions cap_hit1 and 2

src/vec3_linalgebra.c
u413-284-si: True, one needs to take the absolute value of the difference of the two. In case the second float is bigger than the first, the condition would not work as intended anymore. 

Took over the suggestion.

inc/camera.h
u413-284-si: reasonable enough. changed

inc/camera.h
u413-284-si: No the point of having the aspect ratio is to have a fixed ratio and not be defining height AND width which could result into a "weird" ratio. By defining the aspect ratio we only need to determine in our case the image width and the height is calculated automatically.

src/utils.c
u413-284-si: The latter. It shall only be true if the value really is set to 0, nothing else

gwolf-011235: Then no change needed

inc/camera.h
gwolf-011235: Ok

src/hit_cylinder.c
gwolf-011235: Noice!

gwolf-011235 (approved):
Superb!
Everything looks good - can be merged

PR #14, opened 2023-11-20
3 parser

### LINKED ISSUE #3: Parser

gwolf-011235:
Implements a parse Module, which reads in file, performs checks and parses the values.

This fulfills specification and therefore closes #3 

1. Submodule **import_file**: import file and create array of char **lines.
 - Checks if correct data type .rt
 - Opens file
 - Reads in the whole file into a buffer. The initial buffer size can be set (miniRT_config.h). If it is exceeded, the buffer size gets doubled. After the whole file was read ft_split() is called. This gets rid of all newlines effectively ignoring all empty lines. A lines array is created
 - Close the file
 - If any error is encountered print error message and leave module with no allocated memory

2. Submodule **check**: go through lines array and perform format checks
- check if correct line type by comparing identifier (A, C, L, sp, pl, cy)
- if correct line type perform checks according to type. Depending on the type the following checks are performed:
-- ft_isvalid_float: is it a valid float, is it in range (if not specified is set to [-10000, 10000]), is the separator correct 
-- ft_isvalid_float_block: are there 3 valid floats
-- ft_isvalid_rgb_block: are there 3 valid positiv integers in range 0-255
-- every line is checked if it ends with zero to n spaces.
- Valid float check is performed by a test convert using ft_strtod() (see below)
- While performing checks the line is sanitized by getting rid of possible additional spaces.
- If check passes an entity counter is increased and check if maximum number of entities was reached (can be set in miniRT_config.h)
- After all lines were passed check if minimum count of entities ambient, camera and light was reached.
- Set counters for light sources and total of shapes 
-  If any error is encountered print error message and leave module (lines array does not get free'd)
- t_interval was not used, since it would have been harder to implement

3. Module **parse**: call submodules import and check, then parse lines
- If any of the submodules fails, free memory and leave.
- Allocate memory for lights and hittable array
- Parse lines into ambient, camera, lights array and hittable array, keeping track of parsed lights and hittables to correctly access arrays 
- Since lines have been thoroughly checked and sanitized (correct format, every number (block) is followed by exactly one space) parsing can be done without any checks using ft_strtod() (see below)
- Colours are divided by 255.0
- Parsing a hittable sets ID, type and params 
- Cylinder cap1 and 2 are also set

4. Function ft_strtod()
- Used to convert received lines from string to float. 
- Converts number into double, so variables can later be change to double if needed.
- Conversion is oriented on fast convert of David Gay’s strtod(): An integer is created of all digits, and then divided by 10 to the power of the number of digits of the fractional part. Has maximum digits: can only convert 15 digits (indifferent from radix point location), since a variably type long can only savely store that many digits. 
- 123456789012345 works as well as 1.23456789012345, 1234567.89012345 etc.
- Seems like a good compromise: is fast, precise conversion, while also covering a good range.

5. Various
- Error messages are centralized in error_ files
- Comments are supported: if line begins with #, it is ignored.
- Freeing of light and hittable array at end of main needs to be changed when implementing rendering.


Makefile
```diff
@@ -236,20 +201,13 @@ clean:
...
-#	@rm -rf $(TEST)
-#	@printf "$(RED)clean bin $(TEST)$(RESET)\n"
-
 .PHONY: fclean
 fclean: clean
 	@rm -rf $(NAME)
 	@printf "$(RED)clean bin $(NAME)$(RESET)\n"
 	@printf "$(YELLOW)$(BOLD)clean$(RESET) [$(BLUE)libft$(RESET)]\n"
 	@$(MAKE) --no-print-directory -C $(LIB_DIR_FT) fclean
+	@echo
```
u413-284-si, line 210: What's that empty echo for?

inc/check.h
```diff
@@ -0,0 +1,235 @@
...
+t_err		ft_check_lines(char **lines, int *lsrc_count, int *total);
+
+/**
+ * @brief Dispatch function for different line types.
+ *
+ * Searches line for entity identifier.
+ * If found calls and returns correct check function.
+ * If not returns UNKNOWN.
+ * @param line Line to check for identifier.
+ * @return t_type Specific entity type if correct, else UNKOWN.
```
u413-284-si: Misspelled UNKNOWN

inc/entities.h
```diff
@@ -68,41 +68,49 @@ typedef struct s_plane
  * @brief Represents a cylinder
  * @param centre	Centre of the cylinder
  * @param axis		Normalised [-1, 1] axis of the cylinder
- * @param cap1		Centre of first cap = centre - h/2 * axis
- * @param cap2		Centre of second cap = centre + h/2 * axis
  * @param d			Diameter of the cylinder
  * @param h			Height of the cylinder
  * @param colour	Colour of the cylinder
+ * @param cap1		Centre of first cap = centre - h/2 * axis
+ * @param cap2		Centre of second cap = centre + h/2 * axis
```
u413-284-si, line 75: Why you move dis

inc/entities.h
```diff
@@ -68,41 +68,49 @@ typedef struct s_plane
...
+	COMMENT = -2,
+	UNKNOWN = -1,
 	SPHERE,
 	PLANE,
-	CYLINDER
+	CYLINDER,
+	AMBIENT,
+	CAMERA,
+	LIGHT,
+	SUM_ENTS
```
u413-284-si, line 105: Could be misinterpreted as walking tree. How about SUM_TITIES :D

inc/error_msg.h
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @brief Error message if converted number is out of range.
+ *
+ * @param line Start position of converted number.
+ * @param offset Offset from start position.
+ * @param min Minimum for number.
+ * @param max Maximum for number.
+ * @return false Always returns false.
+ */
+bool		ft_perror_range(char *line, size_t offset, int min, int max);
```
u413-284-si: Why have a return value if it is always the same?

inc/error_msg.h
```diff
@@ -0,0 +1,100 @@
...
+ * @brief Error message if wrong separator was encountered.
+ *
+ * @param line Line with current position, where separator was encountered.
+ * @param comma Controls if separator should be ',' or ' ' and '\0'.
+ * @return false Always returns false.
+ */
+bool		ft_perror_separator(char *line, bool comma);
+
+/**
+ * @brief Error message if line doesn't end with space or zero temrinator.
```
u413-284-si: Misspelled terminator

inc/import_file.h
```diff
@@ -0,0 +1,126 @@
...
+ */
+bool		ft_isvalid_filename(char *filename);
+
+/**
+ * @brief Import file and split it into lines.
+ *
+ * Initialize buffer with ft_buf_init().
+ * Read the whole file into buffer with ft_buf_read_file().
+ * Convert file into array of lines with ft_split().
+ * Split char is newline '\n', which gets rid superfluous newlines.
```
u413-284-si: gets rid of

inc/miniRT_config.h
```diff
@@ -0,0 +1,35 @@
...
+/* ===== MACROS ===== */
+
+// init buffer size for file import
+# define BUF_SIZE 4096
+
+// maximum for each entity
+# define AMBIENT_MAX 1
+# define CAMERA_MAX 1
+# define LIGHT_MAX 1
+# define SPHERE_MAX 100
```
u413-284-si, line 25: We don't want more :D?

lib/libft
u413-284-si, line 1: wasn da passiert

u413-284-si: Ok, hab mal drüber geschaut, meist nur comments und silencing von messages im Makefile. Dein ft_mem.h war vorher als standard lib integriert?

scenes/err_empty.rt
u413-284-si, line 1: placeholder for later testing?

u413-284-si: aaah i see, it's an empty file for the parser

scenes/err_hidden_garbage.rt
```diff
@@ -0,0 +1,6 @@
+A     0.2                                          255,255,255     I'm sneaky boy
```
u413-284-si, line 1: mad boi

scenes/minimalist.rt
u413-284-si, line 1: This should be fine or?

src/camera.c
```diff
@@ -23,12 +23,6 @@ void	ft_initiate_image(t_image *image)
 
 void	ft_initiate_camera(t_cam *cam)
 {
-	cam->look_from.x = -0.0;
```
u413-284-si, line 26: When merged my code regarding lookfrom/lookat will be deleted o.O how we prevent this? you add these lines to your code?

src/camera.c
```diff
@@ -43,7 +37,7 @@ void	ft_initiate_camera(t_cam *cam)
 void	ft_initiate_viewport(t_viewport *vp, t_cam cam, t_image image)
 {
 	vp->focal_length = ft_vec3_abs(ft_vec3_sub(cam.look_from, cam.look_at));
-	cam.hfov = ft_degree_to_radian(100);
+	cam.hfov = ft_degree_to_radian(cam.hfov);
```
u413-284-si, line 40: sneaky

src/check_entity_ACL.c
```diff
@@ -0,0 +1,61 @@
...
+/*   Updated: 2023/11/20 09:44:47 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "check.h"
+
+t_type	ft_check_ambient(char *line)
+{
+	ft_rm_space(&line);
+	if (!ft_isvalid_float(&line, 0.0, 1.0, false))
```
u413-284-si, line 18: could have used t_interval

src/check_line.c
```diff
@@ -0,0 +1,96 @@
...
+	char	*begin;
+	char	*end;
+
+	if (**str == ' ')
+		(*str)++;
+	begin = *str;
+	end = *str;
+	while (*end == ' ')
+		end++;
+	while (*end)
```
u413-284-si, line 28: Not sure I understand your purpose here:

If the starting char is a space, you move forward one position. Then you continue moving while you encounter spaces. Once there is sth else than a space you move to the next position and set that to \0? This shouldn't be happening as that would overwrite content of the file. And I don't think that this is the case, so there is sth that I don't see...

src/check_line.c
```diff
@@ -0,0 +1,96 @@
...
+}
+
+bool	ft_isvalid_rgb_val(char **line, bool comma)
+{
+	int	tmp;
+
+	tmp = 0;
+	if (!ft_isdigit(**line))
+		return (ft_perror_convert(*line, false));
+	while (ft_isdigit(**line) && tmp < 255)
```
u413-284-si, line 72: Why not use atoi here?

src/cleanup.c
```diff
@@ -0,0 +1,26 @@
...
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "cleanup.h"
+
+void	ft_free_char_arr(char **array)
+{
+	size_t	i;
+
+	i = 0;
```
u413-284-si, line 19: start with -1 and save three lines ;)

src/error_msg_check.c
```diff
@@ -0,0 +1,89 @@
...
+		ft_putstr_fd("Too many entities of type <", 2);
+	else
+		ft_putstr_fd("Missing entity of type <", 2);
+	ft_putstr_fd((char *)entity_name[type], 2);
+	ft_putendl_fd(">", 2);
+	if (!high)
+		return ;
+	ft_putstr_fd("Expected: ", 2);
+	ft_putnbr_fd(max, 2);
+	ft_putendl_fd("", 2);
```
u413-284-si, line 37: Whats that printing? Nothing?

src/error_msg_check.c
```diff
@@ -0,0 +1,89 @@
...
+	ft_putstr_fd("But got:  ", 2);
+	ft_putnbr_fd(count, 2);
+	ft_putendl_fd("", 2);
```
u413-284-si, line 40: same here

src/error_msg_check.c
u413-284-si: oh i see, it just gives a newline

src/error_msg_check.c
u413-284-si: ignore

src/ft_strtod.c
```diff
@@ -0,0 +1,55 @@
...
+/* ************************************************************************** */
+
+#include "ft_strtod.h"
+
+static size_t	ft_conv_nums(const char *str, long *result)
+{
+	size_t	i;
+
+	i = 0;
+	while (ft_isdigit(str[i]))
```
u413-284-si, line 20: save three lines

src/import_file.c
```diff
@@ -0,0 +1,66 @@
...
+		return (ERROR);
+	if (ft_buf_read_file(&buf, fd))
+	{
+		ft_buf_destroy(&buf);
+		return (ERROR);
+	}
+	errno = 0;
+	*lines = ft_split(buf.str, '\n');
+	if (errno)
+		ft_perror("malloc() failed", errno);
```
u413-284-si, line 48: bit weird but ok. This version is shorter than putting it like 
if (!*lines)
{
ft_perror(blbla);
ft_buf_destroy(&buf);
return (ERROR);
}

Cause of our convention to have each bracket on a new line..

src/import_file.c
```diff
@@ -0,0 +1,66 @@
...
+		return (ERROR);
+	return (SUCCESS);
+}
+
+bool	ft_isvalid_filename(char *filename)
+{
+	int		len;
+
+	len = ft_strlen(filename);
+	if (!ft_strnstr(filename + len - 3, ".rt", 3))
```
u413-284-si: interesting alternative to ft_strncmp

src/import_file_buffer.c
```diff
@@ -0,0 +1,80 @@
...
+	buf->cur_pos = 0;
+}
+
+t_err	ft_buf_double(t_buf *buf)
+{
+	char	*tmp_str;
+	size_t	tmp_size;
+
+	tmp_size = buf->size * 2;
+	if (tmp_size < buf->size)
```
u413-284-si, line 42: How does tmp_size represent the actual available memory space? The limits of size_t are defined independently of the current status of memory. Would leave that check away as available memory would anyhow be determined by the following malloc operation

src/import_file_buffer.c
```diff
@@ -0,0 +1,80 @@
...
+
+t_err	ft_buf_read_file(t_buf *buf, int fd)
+{
+	ssize_t	rd_bytes;
+	char	*cur_str;
+	size_t	cur_read;
+
+	rd_bytes = 0;
+	while (1)
+	{
```
u413-284-si, line 64: Did you recode some form of gnl there??

src/main.c
```diff
@@ -3,14 +3,16 @@
...
+/*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2023/10/16 14:55:47 by sqiu              #+#    #+#             */
-/*   Updated: 2023/11/17 14:28:38 by sqiu             ###   ########.fr       */
+/*   Updated: 2023/11/20 12:33:16 by gwolf            ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
 #include "miniRT.h"
+#include "parse.h"
```
u413-284-si: would put these into miniRT.h

src/main.c
```diff
@@ -21,14 +23,16 @@ int	main(int argc, char **argv)
...
 	}
-	ft_create_scene(&scene);
+	if (ft_parse_file(argv[1], &scene, &cam))
+		exit(1);
 	ft_initiate_image(&image);
 	ft_initiate_camera(&cam);
 	ft_initiate_viewport(&viewport, cam, image);
 	ft_create_image(image, cam, viewport, scene);
-	(void)argv;
+	free(scene.obj);
```
u413-284-si, line 33: Could summarise these two in one function ft_free_all or ft_shutdown so it looks less bare metal

src/parse.c
```diff
@@ -0,0 +1,78 @@
...
+/*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2023/11/12 16:02:54 by gwolf             #+#    #+#             */
+/*   Updated: 2023/11/20 13:03:50 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "parse.h"
+
+t_err	ft_parse_file(char *filename, t_entities *ents, t_cam *cam)
```
u413-284-si: I'd call it scene instead of ents to keep consistency

src/parse.c
```diff
@@ -0,0 +1,78 @@
...
+		return (ERROR);
+	if (ft_err_malloc((void **)obj, sizeof(**obj) * total))
+	{
+		free(*lsrc);
+		return (ERROR);
+	}
+	return (SUCCESS);
+}
+
+void	ft_parse_lines(t_entities *ents, t_cam *cam, char **lines)
```
u413-284-si: I guess this one needs to be shortened at the end

u413-284-si (changes requested):
Nicely done!
I really like the structure, how the whole process is broken down into modules and submodules and subfunctions, one building on the other. Also the trickery with the t_types and using them as index for accessing the numbers in the int array is very neat. All in all very good work! Mostly small remarks, some possible optimisations, code consistency questions and design decisions.

src/parse.c
gwolf-011235: True. Was a last minute change to add comments (#), which i saw in your testfile :D
Problem was that comments would inc the id (since it would fall into else clause). Changed to exclude comments.

Makefile
gwolf-011235: Prints a newline to separate the cleaning step from compilation if `make re` is called. 
I think it looks more neat.
Could also be printf "\n"

inc/check.h
gwolf-011235: fixed

inc/entities.h
gwolf-011235: To be more inline with the .rt file: it gives us info on the first params up until the colour. 
The caps are then calculated by us.

inc/entities.h
gwolf-011235: I'd rather be known as a tolkien fan, than a pervert :O

inc/error_msg.h
gwolf-011235: Because it can then be used in function return chains, i.e. return (ft_perror_range) - in case of error. 
Updated doxy comment: "Always returns false to chain msg and false in return."

src/import_file_buffer.c
gwolf-011235: I decided against the usage of gnl: 
- we would need to protect each gnl call (each call could fail), 
- rewrite it a bit (gnl uses read and malloc so ERRNO could be set with either)

This uses a buffer which just reads in the whole file (so not line by line), making also less read/malloc calls (depending on buffer size). We can also check for fails of read and malloc independently

src/main.c
gwolf-011235: Changed.
Although I think we should - at the end - just include the needed function for isolation

src/main.c
gwolf-011235: Was just temporary to check for leaks with valgrind. I would leave it like that until we know what we really need to clean at the end

src/parse.c
gwolf-011235: Changed

inc/error_msg.h
gwolf-011235: Fixed

inc/import_file.h
gwolf-011235: Fixed

inc/miniRT_config.h
gwolf-011235: Maybe? I just picked a random number 👯‍♀️

lib/libft
gwolf-011235: Got rid of percentage indicator to print the compilation of files - I think this is better since you get more information. Is also more in line with our makefile.
Yeah, ft_mem.h was included with <> which most likely happened while splitting

scenes/err_empty.rt
gwolf-011235: Yes

scenes/err_hidden_garbage.rt
gwolf-011235: he sneaky

scenes/minimalist.rt
gwolf-011235: This is the example file from subject, which represents "a minimalist scene". So yeah should go through - and does

src/camera.c
gwolf-011235: I deleted look_from and look_at since they get parsed. Else you would overwrite the parsed values here. So it should be fine if they also get deleted on your part?

src/camera.c
gwolf-011235: Is just temp, since you pass t_cam by value (not with a pointer)

src/check_entity_ACL.c
gwolf-011235: As i wrote in PR description: i decided against it, since it would be harder to integrate. 
I would need to create a t_interval or pass a compound literal, which usage is still questionable re norm conformity

src/check_line.c
gwolf-011235: We sanitize spaces. 
We move forward one space (line 20, 21) - we want exactly one space as a separator between the parsed infos.
If we then encounter more spaces (line 24, 25) we search the next not space char.
The we copy everything from the point where the spaces end to the point where the first space was, getting rid of the extra found spaces. 
Then we set a new zero terminator.

The line "12.0 &nbsp; &nbsp; &nbsp; 34,56,78" becomes "12.0 34,56,78"

I noticed that I would copy in every case: added a check if begin and end are the same for early return.

src/check_line.c
gwolf-011235: We can't control the endpoint as easily: we want to stop, as soon as we are over 255. With atoi we'd need to check for overflow and if it's under 255.
Also we want to move the line forward for each digit, which means we would need to count again after atoi.

src/cleanup.c
gwolf-011235: I generally use size_t for array iterators, since an array position can never be negative. This version also supports arrays double the size since unsigned

src/error_msg_check.c
gwolf-011235: Yes - no ft_printf to stderr :(

src/error_msg_check.c
gwolf-011235: ok

src/ft_strtod.c
gwolf-011235: As mentioned above: uses size_t

src/import_file.c
gwolf-011235: I would also need the line `ft_buf_destroy(&buf);` twice since I want to destroy in both cases.
Should I change to longer version? We would have space.

src/import_file.c
gwolf-011235: Nice burn! Changed

src/import_file_buffer.c
gwolf-011235: tmp_size is calculated as current size of buffer times two.
If tmp_size is smaller than size of buffer overflow must have happend -> it has to be bigger.
This overflow is catched with this check.

Makefile
u413-284-si: Lo veo

inc/entities.h
u413-284-si: you pencil pusher you

inc/entities.h
u413-284-si: where's the fun in that

inc/error_msg.h
u413-284-si: oke

inc/miniRT_config.h
u413-284-si: Okay, maybe will revisit this

lib/libft
u413-284-si: And it still worked? with the <>

scenes/minimalist.rt
u413-284-si: good

src/camera.c
u413-284-si: Ah, misunderstood then. Thought the two variables get omitted completely. But you still use them in your code to save  the parsed values of camera view point and orientation vector, right? In that case, yeah we can delete these assignments

src/camera.c
u413-284-si: yeah, that's why it's sneaky

src/check_entity_ACL.c
u413-284-si: You can use my t_interval. After merge

src/check_line.c
u413-284-si: Oh you sneaky boi you

src/check_line.c
u413-284-si: Right, the line moving is indeed a handy extra features

src/cleanup.c
u413-284-si: Interesting, but in case we need to save lines I'd still revert to int i = -1. Might not be the case here, but it would be preferable instead of creating a whole new function

src/ft_strtod.c
u413-284-si: as above

src/import_file.c
u413-284-si: Nahh leave it. Just takes a while to grasp what is happening

src/import_file_buffer.c
u413-284-si: You mean overflow of size_t. But what do we care about that? Our only limit is the available memory space

src/main.c
u413-284-si: even better

src/main.c
u413-284-si: I don't think there will be anything else. These are the only two instance requiring mallocing

src/parse.c
u413-284-si: So we don't accept comments in .rt files or how do we handle them?

lib/libft
gwolf-011235: Not sure, never used it I think 🤔

src/camera.c
gwolf-011235: Yes

src/import_file_buffer.c
gwolf-011235: What's the advantage of deleting the line?

src/check_entity_ACL.c
gwolf-011235: I'll try

src/cleanup.c
gwolf-011235: Yes

src/main.c
gwolf-011235: I saw a camera implementation which caches the ray directions - this cache would also need to be freed.
Another malloc would be for mlx. Though this will be contained in the render part I think.
What I want to say: you'll never know whats coming.

But yeah I can move it in its own function

src/parse.c
gwolf-011235: We do: if line starts with # it gets ignored. I just moved some part around to make it fit in 25l

inc/entities.h
gwolf-011235: Me don't care - serous suggestion? :o

lib/libft
u413-284-si: K

src/import_file_buffer.c
u413-284-si: What do you mean? The question we should ask is do we need that line? If no, it shouldn't be there

gwolf-011235: I don't think we're on the same page.

The goal of the function is to double the buffer size.
It gets called when the buffer is full to make room for more data.

Are you saying that malloc should return null if not enough space, making the check not needed?

Malloc - in school version of glibc-  takes in a size_t as size. This means malloc itself is bound by the size of size_t. If I would try to pass something larger it would overflow. I can't malloc a larger block of bytes than size_t.

But let's say that the malloc call is only depending on the available space: If the check before wouldn't check for overflow it could overflow and then call malloc with the reduced size (overflow results in a smaller number than before).

This means that the new malloc buffer size is smaller than before (which is unchecked). Then we would try to write the previous buffer into the new one - which means we lost data (since the new buffer is smaller).

Or are you saying that space would run out before size_t is even reached, because size_t is so big?

Then we rely on an arbitrary memory size and the assumption that memory is always smaller than size_t. Yes if there are not enough bytes we never trigger the check. If there are then we corrupt the input.

Granted there is a very small chance that this even triggers: the file needs to be absurdly big. But it has nearly zero impact on performance (it's called while parsing and even then only if the buffer is full and then it's just compare between two nums), and has a clear function (overflow protection).
I took it from minishell because it's just a general implementation of a buffer double function.

u413-284-si: Ok, now I see what you mean.

You're not protecting against physical buffer overflow (at least with this check), but against your buf->size overflowing which is defined as size_t. While taking into consideration that it is big enough to contain the size of the biggest object the host system can handle, being dependent on the bit size of the underlying OS, that should be a good enough representation of the max capacity of the individual system. Therefore the check makes sense.

u413-284-si (approved):

PR #16, opened 2023-12-19
6 1 renderer

gwolf-011235:
Implements a basis for image rendering in a window via mlx.

Main part is in render.h
- new struct t_img which contains params for a mlx image (potential confusion with existing struct t_image, which contains width, height and ratio for the to be generated image)
- new struct mlx_ptrs which contains mlx pointers and an t_img struct. Last struct could be made into ptr (t_img *img) to support more than one image (not needed right now, maybe for double buffering)
- new struct t_render which combines many structs to pass them around in the mlx loop.

Initialisation of mlx uses error wrappers: if error fails, msg with errno is printed and it returns ERROR. Then everything gets freed and program exits.

If everything is fine the mlx loop gets started. 
- The loop consists mainly of redrawing the image (which doesn't change right now). The image rendering was moved from camera file to render_draw. It draws now on the mlx img which gets then put to screen.
- mlx loop has hooks for ESC key and Window X pressed to end loop and free mlx ressources.
- Another keyhook 'C' which outputs the current image as ppm file. 

New free_scene function to free allocated ressources for the scene (effectively combines two free() lines into one)


inc/camera.h
```diff
@@ -93,40 +93,25 @@ typedef struct s_viewport
...
  * @brief Sets values of viewport parameters
- * 
+ *
  * @param vp	Struct containing viewport parameters
  * @param cam 	Struct containing camera parameters
  * @param image	Struct containing image parameters
  */
 void	ft_initiate_viewport(t_viewport *vp, t_cam cam, t_image image);
 
-/**
```
u413-284-si, line 117: Where'd it goooooo o.O

inc/cleanup.h
```diff
@@ -28,4 +31,20 @@
  */
 void	ft_free_char_arr(char **array);
```
u413-284-si, line 32: Requires documentation

inc/key_codes.h
```diff
@@ -0,0 +1,64 @@
...
+/*   Created: 2023/12/19 22:45:45 by gwolf             #+#    #+#             */
+/*   Updated: 2023/12/19 22:47:15 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#ifndef KEY_CODES_H
+# define KEY_CODES_H
+
+// keyboard
+# define K_ENTER 65293
```
u413-284-si: Can you remind me what the numbers were representing again?

inc/render.h
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @param ptr		Pointer to mlx img.
+ * @param addr		Pointer to the char *buffer of the mlx img.
+ * @param bpp		Bytes per pixel.
+ * @param line_len	Line length.
+ * @param endian	Endian.
+ * @param bytes		Number of bytes per pixel.
+ *
+ */
+typedef struct s_img
```
u413-284-si, line 43: Maybe call it s_mlx_img to avoid confusion with the existing s_image

inc/render.h
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * @param mlx_ptr	Pointer to mlx.
+ * @param win_ptr	Pointer to mlx window.
+ * @param img		mlx img struct.
+ */
+typedef struct s_mlx_ptrs
+{
+	void	*mlx_ptr;
+	void	*win_ptr;
+	t_img	img;
```
u413-284-si, line 66: same here, could call it mlx_img

inc/render.h
```diff
@@ -0,0 +1,171 @@
...
+ *
+ * Loops for the number of bytes per pixel and writes the colour to the address.
+ * @param img	Pointer to mlx img struct.
+ * @param x		X coordinate.
+ * @param y		Y coordinate.
+ * @param color	Colour to write.
+ */
+void	ft_put_pix_to_image(t_img *img, int x, int y, int color);
+
+/**
```
u413-284-si, line 129: Ah I see my ft_create_image() reborn here

inc/render.h
```diff
@@ -0,0 +1,171 @@
...
+ * @brief Renders the scene to the mlx img.
+ *
+ * Creates rays for each pixel of the image and "shoots" them
+ * through the corresponding pixel in the viewport. The colour of
+ * each ray is returned and written to the mlx img.
+ * @param render	Pointer to render struct.
```
u413-284-si, line 135: dont see no pointer

src/cleanup.c
```diff
@@ -24,3 +24,22 @@ void	ft_free_char_arr(char **array)
...
 	free (array);
 }
+
+void	ft_free_mlx(void *mlx_ptr, void *win_ptr, void *img_ptr)
+{
+	if (img_ptr)
+		mlx_destroy_image(mlx_ptr, img_ptr);
+	if (win_ptr)
+		mlx_destroy_window(mlx_ptr, win_ptr);
+	if (mlx_ptr)
```
u413-284-si, line 30: Shouldn't that check encompass the above? As the mlx_ptr is a prerequisite for both functions. Otherwise an error could occur.

Sth like

if (mlx_ptr)
	{
		if (win_ptr)
		{			
			mlx_destroy_window(mlx_ptr, win_ptr);
			win_ptr = NULL;
		}
                if (img_ptr)
		{			
			mlx_destroy_image(mlx_ptr, img_ptr);
			img_ptr = NULL;
		}
		mlx_destroy_display(mlx_ptr);
		free(mlx_ptr);
	}

src/render_keyhook.c
```diff
@@ -0,0 +1,28 @@
...
+
+int	ft_keyhook_press(int key, t_render *render)
+{
+	static bool	print;
+
+	if (key == K_ESC)
+		mlx_loop_end(render->mlx_ptrs.mlx_ptr);
+	else if (key == K_C && !print)
+	{
+		print = true;
```
u413-284-si, line 23: so it doesn't get outputted repeatedly I presume

src/render_loop_mlx.c
```diff
@@ -0,0 +1,35 @@
...
+/*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2023/12/19 22:37:57 by gwolf             #+#    #+#             */
+/*   Updated: 2023/12/19 22:54:05 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "render.h"
+
+int	ft_draw_scene(t_render *render)
```
u413-284-si, line 15: Do we need an extra function for that?

src/render_loop_mlx.c
```diff
@@ -0,0 +1,35 @@
...
+{
+	ft_render_image(*render);
+	return (0);
+}
+
+int	ft_end_loop(void *mlx_ptr)
```
u413-284-si: Sames as above

u413-284-si (changes requested):
Nicely done!

No major issues, variable naming of t_img could be revisited to avoid confusion. Other than that only small changes required.

Integration of my part will be a pain though as some significant refactoring happened xD

inc/camera.h
gwolf-011235: Like I wrote in PR description:
"The image rendering was moved from camera file to render_draw."

Makes more sense to me since the Camera does sees things but does not render.

inc/key_codes.h
gwolf-011235: After establishing a connection to the Xserver window (via minilibx), the window can notify us about different events, like mouse movement over the window, or that a key was pressed.
With `mlx_loop_hook` we can listen to those events, and act accordingly.

The macros are for readability: a KeyPress event reports that a specific key was pressed, but we only receive a corresponding integer (e.g. if Escape is pressed we receive 65293). 
With the macros we can write `if (key == K_ESC)` 

The Codes start with a K_ for keyboard and M_ for mouse.

inc/render.h
gwolf-011235: I think we can get rid of t_image. 
We only need it for mlx at startup (since the size of the image buffer after mlx init doesn't change). 
And we need it for viewport init (and later viewport update). But I'd rather save the info in t_viewport itself. Right now its a bit clunky since we need to pass both structs.
Or we save the params in t_img, since t_image precisely represents the image which is written to by mlx (it has the height and width of the image). These exact params are used when creating the mlx img.
I think it would be more confusing to have them both since its not exactly clear, what the difference between those two is.

inc/render.h
gwolf-011235: Yes

inc/render.h
gwolf-011235: This was an oversight, but also addresses a spicy topic: is there a difference between passing by value or passing by pointer?

The highest rated answer on SO thread from 2020 says with no optimizations it should be the same. But with optimizations it makes a difference: [Link to Thread](https://stackoverflow.com/questions/59976295/structs-and-unions-which-is-better-from-a-performance-point-of-view-passing-th)
Interestingly enough another one argues that the performance gain is negible on newer machines, so that it only makes a difference on embedded or mobile.

On the ACCU21 conference there was a talk about Modern C: [Link to YouTube Video](https://youtu.be/QpAhX-gsHMs?si=Eb_RAOE4TbBZdKhV)
The speaker states, that passing by value is generally more efficient nowadays, than passing by pointer (which contradicts the thread above).

Right now we init t_vec3 structs via value return.
We generally pass pointers of structs to functions but also sometimes not (e.g. ft_initiate_viewport() )
I wouldn't touch the t_vec3 fts since they are nice to work with. But I think we should uniformly pass by value or pointer 

A pointer is 8byte in size and our structs are all bigger than that. Therefore pointer passing is generally lighter. Going by SO passing by pointer should be better with optimizations.
Since in base we're not "allowed" to compile with optimizations - at least for base subject - do we care about optimizations on this level? 
The advantage of always passing by value would be that it is easier to write (render.cam instead of render->cam).

src/cleanup.c
gwolf-011235: If mlx_ptr doesn't exist, then win_ptr and img_ptr cannot exist since the creating functions need to be passed a mlx_ptr.
So the error cases are
- mlx_ptr was free'd and set to NULL before freeing the other stuff -> we won't do that as evidenced by this function. 
- mlx_ptr was "lost" -> at this point we would crash anyways, before we could free something.

src/render_keyhook.c
gwolf-011235: Yes, if pressed again while the function prints, the file gets corrupted

src/render_loop_mlx.c
gwolf-011235: Placeholder for now. Plan is to also render a menu (for example a bar on the left side) with info about the current object which would also be called here. 
Although the names should be flipped, since scene is the thing which gets raytraced.

src/render_loop_mlx.c
gwolf-011235: I think I had a problem with passing the function to the loop hook, but it should work. Will check!

inc/render.h
gwolf-011235: See above

inc/cleanup.h
gwolf-011235: Will add

inc/camera.h
u413-284-si: yeah I saw it later on

inc/key_codes.h
u413-284-si: The numbers are randomly chosen/predefined by minilibx?

inc/render.h
u413-284-si: If we can fuse them into one, the better. Especially if the same variables are used in both

inc/render.h
u413-284-si: I would go with pointers as this should be in theory more efficient, especially if the content of that pointer supersedes it by far in size. Also it keeps the passed size nicely uniform.

It may not make a difference in the base version, but then again, does it really make a difference at all? One that we would notice? On our machines the calculation power is that immense that I doubt that the jump from address to value results in a noticeable difference..

src/cleanup.c
u413-284-si: You never know what a handy user could do ;) sneakily freeing mlx_ptr while prog is running for example... would always go with the safer version. It's not like we're short on lines here

src/render_loop_mlx.c
u413-284-si: I see

inc/key_codes.h
gwolf-011235: Kinda: The numbers are set by the X11 library, around which minilibx wraps. 
65293 may look nicer in hex: 0xff1b. It's only so high because it's a "special key". For example C has 0x63, which is 99, which is also it's value on the ASCII table.

I found out, that in context of X11 there is a difference between keycodes and keysymbols. 
- Keycode represents the physical hardware location: the first key, on the top row
- Keysymbol represents the higher layer: which function has a key, e.g. print C.

I'll rename the file.
The keys and symbols are not standardized. The "real" way of determining the key values and symbols would be to use a tool like xev or in programming the function XLookupKeysym(). Both we cannot do. But on the school machines as well as my WSL-Ubuntu, the keymappings are correct.
There is also X11/keysemdef.h which has the same Macros defined, but if I include it it doesn't find the macros! Maybe you get it to work?

inc/render.h
gwolf-011235: Will do

inc/key_codes.h
u413-284-si: We can certainly try. Can bring my notebook tmr to campus

inc/render.h
gwolf-011235: Maybe it was not clear: I don't know which one is better, since the two sources say the exact opposite. 
Passing by pointer should be faster would also be my first intuition since we don't need to allocate the whole struct on the stack. 
However passing by value seems to be highly optimized anyway --> maybe its only true for newer standards like C11 or C20?

I think we shouldn't go for which is faster but which is more readable.
And here passing by value wins in my book, since we don't need the arrow operator. This makes it shorter and for me more readable. Even more so when a subelement is addressed:

`render.cam.look_at` vs `render->cam.look_at`

u413-284-si: I don't see no issue in readability with the pointers. 
Regarding efficiency: if there is a debate going on between the fractions, we are certainly not (yet) in a position to find a solution. So it will be more a gut decision, and my gut also tells me that pointers should be better.

But then again, I don't think it really matters

gwolf-011235: I updated ft_draw_scene and initiate_viewport.

I checked and you'd need to update most of the raytracer functions. Almost all of them pass by value (ray.h, lighting.h, hit_manager.h). Currently pass by pointer is only used if you want to change the thing in the function.

src/cleanup.c
gwolf-011235: How would the user do that? :o 

And even if he can: is it really expected, that a program does not crash if someone randomly sets a pointer to NULL? Following this argument: Shouldn't we then check every time we use a allocated ressource, since a handy user could set it to NULL? What is the difference with ressources allocated by mlx?

The function gets called while init_mlx and after loop end.
While init_mlx the function call gets filled progressively: 
- if mlx_ptr is success, and win_ptr fails, the mlx_ptr is passed and the other two args are set to NULL
- If mlx_ptr and win_ptr is success, but img ptr fails we pass two pointers, set one to NULL. 
- if all three are success we are happy and don't free anything, since we can get to work.

The function gets called again right after loop end to clean up. While looping we need the mlx_ptr to maintain the XServer connection, else we crash. After loop end we free the three ressources. Since all of them are needed for our program to enter the loop, all of them should be there.

I don't think it introduces more safety, but it just a more contrieved way of doing the same thing.

inc/key_codes.h
gwolf-011235: Nevermind, I just forgot to include the Xlib.h header. Now we can use the definitions from keysymdef.h directly :D

src/cleanup.c
gwolf-011235: Changed to suggested version

u413-284-si (approved):
gg

PR #18, opened 2023-12-23
6 2 renderer

gwolf-011235:
Second part of renderer

### Adds more keys to manipulate objects (hittables)

- Change active hittable: N M
- Movement: x (AD) y (WS) z (QE)
- Radius: bigger (R) smaller (F)
- Rotation: Arrow keys, X-rotation (left, right) Y-rotation (up, down)
- Colors: number keys 1 to 0
- Is not displayd anywhere right now - would be in menu

### Debug print

- Functions to print values of vec3 or colour or whole objects 

### Makefile update

- Different obj subdirectories and binary name for different builds (default, leak, address, speed)
- make is now outputs less per default but can be made verbose if variable VERBOSE is set 
- simple counter to see how many files changed


Makefile
```diff
@@ -27,12 +38,30 @@ BLUE := \033[34m
...
+# Set default object directory
+OBJ_DIR = $(OBJ_DIR_DEFAULT)
+# Set object directory depending on target
+ifneq (,$(findstring leak,$(MAKECMDGOALS)))
+	OBJ_DIR = $(OBJ_DIR_LEAK)
+else ifneq (,$(findstring address,$(MAKECMDGOALS)))
+	OBJ_DIR = $(OBJ_DIR_ADDRESS)
+else ifneq (,$(findstring speed,$(MAKECMDGOALS)))
+	OBJ_DIR = $(OBJ_DIR_SPEED)
+endif
```
u413-284-si, line 60: You like playing around with Makefile huh

Makefile
```diff
@@ -130,25 +183,29 @@ all: $(NAME)
...
 	@printf "$(BOLD)$(GREEN)$(NAME) created!$(RESET)\n\n"
 
-# This target adds fsanitize leak checker to the flags. It needs to clean and recompile.
+# This target adds fsanitize leak checker to the flags.
 .PHONY: leak
 leak: CFLAGS += -fsanitize=leak
 leak: LDFLAGS += -fsanitize=leak
-leak: clean $(NAME)
-	@printf "Compiled with $(YELLOW)$(BOLD)fsanitize=leak$(RESET)\n\n"
+leak: $(NAME)
```
u413-284-si, line 194: No more cleaning necessary?

Makefile
```diff
@@ -130,25 +183,29 @@ all: $(NAME)
...
 
-# This target adds fsanitize address checker to the flags. It needs to clean and recompile.
+# This target adds fsanitize address checker to the flags.
 .PHONY: address
 address: CFLAGS += -fsanitize=address
 address: LDFLAGS += -fsanitize=address
-address: clean $(NAME)
-	@printf "Compiled with $(YELLOW)$(BOLD)fsanitize=address$(RESET)\n\n"
+address: $(NAME)
```
u413-284-si, line 200: As above

Makefile
```diff
@@ -130,25 +183,29 @@ all: $(NAME)
...
+
+# This target adds flags which optimize the program for speed.
+.PHONY: speed
+speed: CFLAGS = -Ofast -march=native -fomit-frame-pointer
```
u413-284-si: -fomit-frame-pointer is already included in -0fast

    Omit the frame pointer in functions that don’t need one. This avoids the instructions to save, set up and restore the frame pointer; on many targets it also makes an extra register available [...]

    Enabled by default at -O1 and higher.

Makefile
```diff
@@ -165,25 +222,33 @@ valgr: $(NAME)
...
 # 					so that it will be created when needed.
+# $(eval ...) =		Increment file counter.
 # $(POSTCOMPILE) =	Move temp dependency file and touch object to ensure right timestamps.
 $(OBJ_DIR)/%.o: $(SRC_DIR)/%.c $(DEP_DIR)/%.d message | $(DEP_DIR)
-	$(COMPILE) $< -o $@
-	$(POSTCOMPILE)
+	$(eval CURRENT_FILE=$(shell echo $$(($(CURRENT_FILE) + 1))))
+	@echo "($(CURRENT_FILE)/$(TOTAL_FILES)) Compiling $(BOLD)$< $(RESET)"
+	$(SILENT)$(COMPILE) $< -o $@
+	$(SILENT)$(POSTCOMPILE)
```
u413-284-si, line 240: You reaally like playing with it

inc/colour.h
```diff
@@ -72,4 +86,12 @@ int32_t		ft_convert_colour2int(t_colour colour);
  */
 t_colour	ft_convert_int2colour(int colour);
 
+/**
+ * @brief Returns a colour based on the given colour name.
+ *
+ * @param col_name	Colour name.
+ * @return t_colour	Colour.
+ */
+t_colour	ft_get_colour(t_col_name col_name);
```
u413-284-si, line 95: I already have a function called
t_colour	ft_get_colour(int iterate[2], t_viewport vp, t_cam cam, \
	t_entities scene)
for the anit-aliasing :D

src/colour.c
```diff
@@ -64,4 +64,30 @@ t_colour	ft_convert_int2colour(int colour)
 	ret.g = colour >> 8 & 0xFF;
 	ret.b = colour & 0xFF;
 	return (ret);
-}
\ No newline at end of file
+}
+
+t_colour	ft_get_colour(t_col_name col_name)
```
u413-284-si, line 69: As mentioned before, needs a different name

src/mat4.c
```diff
@@ -0,0 +1,73 @@
...
+/*   Updated: 2023/12/20 11:35:44 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "mat4.h"
+#include "ft_print.h"
+
+t_mat4	ft_mat4_set_identity(void)
+{
+	static const t_mat4	identity = {
```
u413-284-si, line 18: How do you guarantee that the other elements are 0?

src/mat4.c
```diff
@@ -0,0 +1,73 @@
...
+			{
+				ret.mat[j][i] += left.mat[k][j] * right.mat[i][k];
+				k++;
+			}
+			j++;
+		}
+		i++;
+	}
+	return (ret);
+}
```
u413-284-si, line 73: so many brackets :o

src/parse.c
```diff
@@ -24,6 +24,10 @@ t_err	ft_parse_file(char *filename, t_entities *scene, t_cam *cam)
 		ft_free_char_arr(lines);
 		return (ERROR);
 	}
+	if (scene->total != 0)
```
u413-284-si, line 27: What's that do?

src/render_keyhook_hittable.c
```diff
@@ -0,0 +1,83 @@
...
+	ft_print_hittable(*hittable);
+}
+
+
+void	ft_manip_sphere(int key, t_sphere *sp)
+{
+	if (key >= XK_0 && key <= XK_9)
+		ft_keyhook_change_col(key, &sp->colour);
+	else if (key == XK_r)
+		sp->r -= MV_UNIT;
```
u413-284-si, line 51: needs to be changed to sp->d, it was a bug in the previous code. We are given the diameter

src/render_keyhook_hittable.c
```diff
@@ -0,0 +1,83 @@
...
+void	ft_manip_cylinder(int key, t_cylinder *cy)
+{
+	if (key >= XK_0 && key <= XK_9)
+		ft_keyhook_change_col(key, &cy->colour);
+	else if (key >= XK_Left && key <= XK_Down)
+		ft_keyhook_rot_vec(key, &cy->axis);
+	else if (key == XK_r)
+		cy->d -= MV_UNIT;
+	else if (key == XK_f)
+		cy->d += MV_UNIT;
```
u413-284-si, line 77: We need to be able to change the height as well according to subject

u413-284-si (changes requested):
Good stuff! Not much to be altered or added. The Makefile features are a blast!

Makefile
gwolf-011235: Makes sense in this project since we want different versions (debugging, speedy boy) :D

Makefile
gwolf-011235: No: We have different subdirectories for the differently compiled objects. E.g. if you compile "normal" objects (and depends) are in obj/default, if you compile for speed it gets put in obj/speed.
This increases the need for disk space (since we store both compiled versions of objects) but also makes compile times much faster if switching between targets

Makefile
gwolf-011235: No

Makefile
gwolf-011235: True! Deleted

Makefile
gwolf-011235: It was to much output :o

Makefile
u413-284-si: I see, so we do not need to recompile when different flags are set

src/mat4.c
gwolf-011235: The C99 standard guarantees it:
6.7.8 paragraph 19 "[...] all subobjects that are not initialized explicitly shall be initialized implicitly the same as objects that have static storage duration."
And 6.7.8. paragraph 10: "[...] If an object that has static storage duration is not initialized explicitly,
then: [...] if it has arithmetic type, it is initialized to (positive or unsigned) zero;"

src/mat4.c
gwolf-011235: Readable!

src/parse.c
gwolf-011235: If we don't have any hittables - which is a correctly formatted .rt file - we want to init the active hittable to -1 to return when we try to change it

src/mat4.c
u413-284-si: So whenever an arithmetic variable is defined static, it is initialised to zero?

src/mat4.c
u413-284-si: naaah, I find the brackets confusing, more stuff that distracts

src/render_keyhook_hittable.c
gwolf-011235: I think it makes more sense to leave it and change it when you merge your fix: I think it's more than just swapping the letter from r to d?

u413-284-si: In the calculation of the hit of the sphere the code needed to be adjusted, but that's the only instance next to of course the parsing. I already did that. As you want, we can wait.

src/render_keyhook_hittable.c
gwolf-011235: Implemented with keys T and G

src/mat4.c
gwolf-011235: Yes 
- if you have `static int count`, it is init to 0
- if you have `static bool switch` it is init to false (since false is 0)

But: since they are static any subsequent function call will have the previous value. 
If you have a function which you only call once or have a const static value (as seen here) you can use it to init to zero. Otherwise strange bugs may happen

u413-284-si: Good

inc/colour.h
gwolf-011235: I think your functions does a whole lot more than just get a colour :O

src/colour.c
gwolf-011235: I disagree: this gets colour, identified by colour name. Getter functions are a known concept, which return often simple things.
Yours seems to do a whole lot more, judging from the params alone.
Maybe ft_calc_antialiased_colour() or ft_calc_aa_colour() if you shorten antialiasing

u413-284-si: Alright alright, I can change my name :P I'll think of sth nice

u413-284-si (approved):

u413-284-si:
We good to merge?

gwolf-011235:
Yeah should be good!

PR #21, opened 2023-12-25
Horizontal tab recognized while file checking

### LINKED ISSUE #19: Whitespaces in rt files
Currently it seems that only spaces are accepted between number blocks

Whenever a tab or another kind of whitespace is present, the prog throws an error à la: 

Error
Line type not recognized
Line number [ignoring empty lines]: <0>

At least tabs should be accepted

gwolf-011235:
### New separator accepted: horizontal tab '\t'
- Values in .rt files can be separated with space or horizontal tab between numbers or blocks. 
- If horizontal tab is encountered it is set to space for parsing. Additional spaces or tabs are removed, like before.

Closes #19


src/check.c
```diff
@@ -40,19 +40,19 @@ t_err	ft_check_lines(char **lines, int *lsrc_c, int *total)
 
 t_type	ft_check_line_type(char *line)
 {
-	if (!ft_strncmp(line, "A ", 2))
+	if (*line == 'A')
```
u413-284-si, line 43: Shall we also accept lower case letters? Encountered some .rt files with a for ambient for example

src/check_line.c
```diff
@@ -19,9 +19,14 @@ void	ft_rm_space(char **str)
 
 	if (**str == ' ')
 		(*str)++;
+	else if (**str == '\t')
+	{
+		(*str)++;
+		*(*str - 1) = ' ';
```
u413-284-si, line 25: sneaky

u413-284-si (commented):
Not much to be said, quite straightforward

src/check.c
gwolf-011235: I'd say no: it is an identifier therefore case matters. 
Also the hittables all have two lowercase letters which makes for a cool distinction: Uppercase is stuff around the scene; lowercase is stuff in the scene (hittables/objects).

src/check_line.c
gwolf-011235: Gets the job done :D

src/check.c
u413-284-si: If you feel strongly about it, we can keep it this way. For me it would have been just more convenient to be able to just copy the file and run it, instead of having to change the identifiers to upper case.

Had similar issues with other repos where the parse/check was so strict that it didn't accept files which would run under other progs. These were often also not very clear about what blocked them. Finally just discarded them as they were too user unfriendly.

gwolf-011235: I think it just opens up some space for potential bugs (on the top of my head 'C' and 'cy' could be confused) for not much upside. 
On the flip side a search replace for a > A could make these files working

u413-284-si: Does not work for c > C tho. Would turn our cy into Cy. It remains a pain in the ass :(

gwolf-011235: In VSCode: Find and replace (Ctrl+H) with Match Case (Alt+C) and Match Whole Word (Alt+W) should do the trick

u413-284-si (approved):

PR #22, opened 2023-12-25
6 3 renderer

### LINKED ISSUE #6: Renderer
The colour information returned by the ray tracer is to be outputted on the screen with the help of minilibX using the X protocol, thus creating a window containing the image rendered.

The generated window shall have following properties:
- fluid window display (not affected by minimising, overlaps with other windows)
- able to adjust resolution
- register events
   * Closing with Esc & x-button
   * Translation (objects, light, camera)
   * Rotation (objects {except for sphere}, camera)
- render by putting an image, not individual pixels
- error check while initialising
- free all allocated resources upon closing (destroy window, display, free mlx_ptr)

### LINKED ISSUE #12: Change FOV from degree to radian
As mentioned here:

vfov changed to hfov. will be converted to radian while parsing

_Originally posted by @u413-284-si in https://github.com/u413-284-si/miniRT/pull/11#discussion_r1397322246_

gwolf-011235:
Last part of renderer

### Manipulation of camera
- Camera movement is tied to left click of mouse: a click (press and release) toggles camera mode
- Move camera along direction vector (WS), along right vector (AD) or along world up (QE)
- Rotate camera in x and y
- Change field of view ("zoom") with RF
- If right button is pressed the camera can also be rotated (needs to be hold down)

### Revamp of camera and viewport
- Different parts of camera need to be updated when different things happen (movement, rotation, fov). Re-init camera and viewport everytime would be wasteful.
- Split the init_cam and init_viewport to accomodate for this.
- Only specific values need to be stored in cam and viewport struct, precisely those which are needed for later calculations. Superfluous variables removed and only used as stack variables.
- Viewport struct was renamed to pixel grid to more accuratly describe what it does.
- Cam also stores the image size. This could be different than window size, depending on menu implementation. Is needed for viewport dimensions and pixel grid recalculations

### New vectors
- vec2f and vec2i for dimensions in float or int respectively
- Often times x and y coordinates are needed, array (int size[2]) was a bit too clunky for return value.

### New make target
- `make profile` creates profile data (interesting: in ft_vec3_scale most time is spent)
- logs are saved in subdirectory log 
- same as for valgrind: `make valgr`

Closes #12 : hfov is saved in radian
Closes #6 : most parts should be implemented now


inc/camera.h
```diff
@@ -15,83 +15,146 @@
...
+	t_pixel_grid	pixels;
 }	t_cam;
 
+/* ====== FUNCTIONS ====== */
+
+/**
+ * @brief Initializes camera parameters.
+ *
+ * Calculates the base vectors.
+ * Sets the focal length to 10.0.
```
u413-284-si: The focal length is derived from the length of the difference between the look_from and look_at vector

inc/camera.h
```diff
@@ -15,83 +15,146 @@
...
+ * Sets the image size to the given size.
```
u413-284-si, line 76: Where is image size given?

inc/camera.h
```diff
@@ -15,83 +15,146 @@
...
+ * @param img_size	Width and height of image in pixels
+ */
+void	ft_cam_init(t_cam *cam, t_vec2i img_size);
+
 /**
- * @brief Contains viewport parameters
+ * @brief Calculates basis vectors u, v, w
+ *
+ * Calculates the orthonormal basis (coordinate system) to describe the
+ * cameras orientation. The camera is looking in the direction of the
```
u413-284-si: The vector w is customarily defined as opposite to the camera view direction.

src/camera.c
```diff
@@ -6,43 +6,50 @@
...
-	cam->vup.x = 0.0;
-	cam->vup.y = 1.0;
-	cam->vup.z = 0.0;
-	cam->camera_centre.x = cam->look_from.x;
-	cam->camera_centre.y = cam->look_from.y;
-	cam->camera_centre.z = cam->look_from.z;
-	cam->w = ft_vec3_norm(ft_vec3_sub(cam->look_from, cam->look_at));
-	cam->u = ft_vec3_cross(cam->vup, cam->w);
+	ft_cam_calc_base_vectors(cam);
+	cam->focal_length = 10.0;
```
u413-284-si: See before

src/camera.c
```diff
@@ -6,43 +6,50 @@
...
+	cam->image = img_size;
+	ft_cam_calc_viewport_dimensions(cam);
+	ft_cam_calc_pixel_grid(cam);
+}
+
+void	ft_cam_calc_base_vectors(t_cam *cam)
+{
+	const t_vec3	world_up = {0.0, 1.0, 0.0};
+
+	cam->w = ft_vec3_norm(ft_vec3_scale(cam->direction, -1));
```
u413-284-si: That's how the base vectors are calculated after investigating different sources and repos:
	
cam->vup = (t_vec3){0.0, 1.0, 0.0};
	cam->camera_centre = cam->look_from;
	cam->look_at = ft_vec3_add(cam->camera_centre, cam->camera_direction);
	cam->w = ft_vec3_norm(ft_vec3_sub(cam->look_from, cam->look_at));
	if (cam->camera_direction.y == 1)
		cam->u = (t_vec3){1, 0, 0};
	else if (cam->camera_direction.y == -1)
		cam->u = (t_vec3){-1, 0, 0};
	else
		cam->u = ft_vec3_norm(ft_vec3_cross(cam->vup, cam->w));
	cam->v = ft_vec3_cross(cam->w, cam->u);

This is used currently in my code.

u413-284-si: camera_direction is the orientation vector parsed from file

src/camera_movement.c
```diff
@@ -0,0 +1,30 @@
...
+/*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2023/12/24 22:38:49 by gwolf             #+#    #+#             */
+/*   Updated: 2023/12/24 22:55:23 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "camera.h"
+
+void	ft_cam_move_forward(t_cam *cam, float distance)
```
u413-284-si, line 15: Is distance not a constant value set by us/in the config file? Could just use a macro and give the direction with a bool.

src/parse_entity_ACL.c
```diff
@@ -20,9 +20,10 @@ void	ft_parse_ambient(char *line, t_light *ambient)
 
 void	ft_parse_camera(char *line, t_cam *cam)
 {
-	ft_parse_float_block(&line, &cam->look_from);
-	ft_parse_float_block(&line, &cam->look_at);
+	ft_parse_float_block(&line, &cam->position);
```
u413-284-si: So the first parse into position is ok.
The second one can be parsed into a variable called direction, but then in the camera initiation, this direction must be added to position to arrive at look_at. Then with look_at and look_from(=position) we can derive the base vector w and the focal length. See previously posted code.

u413-284-si (changes requested):
It's nice that we can control the cam with the mouse now! Also the implementation of the matrices is good as it seems that I will need them and the belonging operations for a cleaner hit calculation, especially when introducing the 4th object now.

As hinted at, the camera needs to be revisited as my research brought some new insights to light. 

The printing functions serve for the coming menu I suppose?

inc/camera.h
gwolf-011235: We don't exactly have look_at. 
We get the directional vector, which could be interpreted as point, which we look at. But this vector could be the same as look_from, which would result in (0,0,0) after substraction, needing an extra check. The question then becomes: where do we look at if they are the same?

Our setup is more akin to what is hinted at Raytracing in One Weekend 12.2: we have a direction defined in which we look.
The position and the direction can be the same as they are independent from another. This is also why I renamed it to position and direction.

Right now the focal length is therefore just arbitrarily set.

We could evolve the focal length into focus distance as described in Raytracing in One Weekend 13: defocus blur

inc/camera.h
gwolf-011235: Right now it is fixed as macro.

It gets set while initialisation of mlx and then passed to the camera to also save it there.

The goal is to have the possibility to influence the image size, but we need to think about how we want to do it. 
I have it at the back of my mind, but I could also open an issue so we don't forget

inc/camera.h
gwolf-011235: Which is what happens: 
`cam->w = ft_vec3_norm(ft_vec3_scale(cam->direction, -1));`

One could argue that ft_vec3_norm is needless, since it should be normalized anyway.

But good point: why is it defined that way? 

We parse the directional vector, which we then flip  to get w. 

The difference is that if we have position (0,0,1) and direction (0,0,1) we would see the point (0,0,0) even though we look in positive direction  and are positioned after this point.

I was hesitant to delete the line (we could just not flip the vector). But I don't understand why we would need to do it (for me it's a needless operation)

src/camera.c
gwolf-011235: Some of these operations cancel each other out.

We want to calculate the local coordinate system of the camera, which are three orthonormal vectors - essentially x, y and z of the camera, or direction, up and right. We are fancy, so we call them u, v, w. 

Right now it is calculated as:
```
1 const t_vec3	world_up = {0.0, 1.0, 0.0};

2 cam->w = ft_vec3_norm(ft_vec3_scale(cam->direction, -1));
3 cam->u = ft_vec3_cross(world_up, cam->w);
4 cam->v = ft_vec3_cross(cam->w, cam->u);
```
I would even get rid of the flipping, since I don't understand why we do it. This would mean we parse directly into cam.w.

The code above does the same thing, except that it guards against a special case (good) and does some not needed operations (bad). Simplified:

- Define world_up as (0,1,0)
- A = A > Rename the camera center vector to look_from. I think we should use one nomenclature and stick to it, not mix around. This could also mask some stuff which is happening (as demonstrated right here)
- C = A + B > Calc the look_at vector. Since it is adding, we at least don't run into the (0,0,0) issue.
- w = A - C == A - (A + B) == -B > we did some calculations to set w to flipped direction
- Check if cam is looking exactly up or down. In this case we would have a problem since 
(0,1,0) x (0,1,0) == world_up x cam_direction = (0,0,0), which is no unit vector
- Cross cam_direction and cam_up for cam_right

Line 1 saves the world_up vector in cam struct. We generally only need it when calc cam_up, which is why I have it as static const in line 1 of the function.
Lines 2-4 are obfuscating what is happening, they are essentially the same as 2.
The protection against world_up == direction is good. I will add that.
The last line is the same.

src/camera_movement.c
gwolf-011235: Right now yes. 

The idea is to at least be able to somewhat influence the value. 
E.g. in my fdf I had the option to set the incrementer to min/mid/max. If I then zoom, it would add little/somewhat/much to the zoom value. Same thing could be applied here: right now we only add 1. We could influence this value to make the user experience more satisfying.

src/parse_entity_ACL.c
gwolf-011235: Like above: I think we would just do some useless operations.

gwolf-011235:
> It's nice that we can control the cam with the mouse now! Also the implementation of the matrices is good as it seems that I will need them and the belonging operations for a cleaner hit calculation, especially when introducing the 4th object now.
> 
> As hinted at, the camera needs to be revisited as my research brought some new insights to light.
> 
> The printing functions serve for the coming menu I suppose?

Quaternions or rotors would even be faster than matrices :O

I think the camera is essentially calculated the same albeit a additional check is added.

No the printing functions are for debugging purposes, for the menu we need to print to the window.

inc/camera.h
u413-284-si: Exactly, we get a direction and basically  the focal length is the length of this direction vector. 

Our original interpretation of look_at = direction (= vector parsed) was faulty. In that case we'd have to deal with the case you described with a focal length of 0. 

But we don't as Look_at is derived as follows:
look_at = look_from + direction (ref: RayTracing in One Weekend, Fig. 19)
Therefore we will always have a focal length of actually one, given that the parsed direction vector is supposed to be normalised.

inc/camera.h
u413-284-si: In the past it was part of the .rt file with 

R 800 600

for example. In the base part we can just leave it as configurable macros. In the bonus part we could simply parse it in our variables.

inc/camera.h
u413-284-si: The w vector is opposite to view direction due to the right-hand convention for an ortho-normal base. 

When we look into the scene we want u to be the vector pointing right, v pointing upwards. Due to the convention mentioned above, the third vector w is pointing towards us, therefore opposite to view direction. It has been defined as such in mathematics who knows when, but a definition had to be set at one point in time in order to work from there as a standard and reference.

In your example when we sit at (0,0,1) and look into (0,0,1) meaning our viewport aka lookat lies at (0,0,2), we would only see objects with positive z-values. Anything behind us will not be visible (except for we rotate the cam). The fact that w points in the other way does not mean that we also see in that direction. It is just a reference for further calculations.

src/camera.c
u413-284-si: As explained above, that line should be adapted to either

cam->focal_length = ft_vec3_abs(ft_vec3_sub(cam->look_from, cam->look_at));

or even faster

cam->focal_length = ft_vec3_abs(cam->direction);

src/camera.c
u413-284-si: Let's stick to look_from and look_at, these seem to be commonly accepted terms in the raytracing field.

The vup vector can/should never be the same as w. The cases when the cam is looking up or down (cam->direction = 1 or -1) are covered by the extra cases. Any other position of the camera is then calculated with the cross product.

We can leave vup as a const.

src/camera_movement.c
u413-284-si: You mean one key which increases and another which decreases movement speed?

distance = MACRO_VALUE - INCREMENT x keystroke

Sth like that?

src/parse_entity_ACL.c
u413-284-si: Per further explanations above. If need we can talk in person next week.

u413-284-si:
Yeah, let's just try to finish the project rn. It is time...

We can add features and improve performance after validation cause there is still so much that could be done. But every feature takes up so much time to implement, e.g. the bump mapping apparently requires parsing a normal map which is an additional file in itself and then this needs to be processed to be adapted to the UV coordinates of our objects in order to alter the normal vectors.. much more complicated than initially expected. I still have no clear idea of the actual implementation steps but it seems the scope of work is fairly more than a couple of days.
Therefore we should stick our heads together and decide on a timeline when we want to finish (the base part) and what we want/can include in it considering the time costs of each feature still open:

* Menu
* Multi-Threading
* Cone (on it since two days, not happy)
* Textures (colour disruption, bump mapping)
* Further optimisations...

inc/camera.h
gwolf-011235: So just set focal length to 1?

> But we don't as Look_at is derived as follows:
> look_at = look_from + direction (ref: RayTracing in One Weekend, Fig. 19)

This has the same limitations as substraction just with opposite signs:
look_from (0,0,1) and direction (0,0,-1) leave us with (0,0,0) as look_at.

We need to evolve from the base part of the book. In the book the cam is fixed at (0,0,0) which makes it easy to use look_at. We move the cam, so we should be more general (which is easy with the given direction vector)

inc/camera.h
gwolf-011235: OK, then we finished for base!

src/camera.c
gwolf-011235: Or just cam->focal_length = 1 since its normalized?

src/camera.c
gwolf-011235: > Let's stick to look_from and look_at, these seem to be commonly accepted terms in the raytracing field.

Regarding naming? What about the (in my view needless) calculations? 

> The vup vector can/should never be the same as w. The cases when the cam is looking up or down (cam->direction = 1 or -1) are covered by the extra cases. Any other position of the camera is then calculated with the cross product.

Yes it can be the same, which is why we need the special cases.

src/camera_movement.c
gwolf-011235: Pass as distance a variable which is manipulated by a keypress.

I would call it like ft_cam_move_forward(&render->cam, render->increment), where increment can be manipulated by a keypress to be either 1, 2 or 3 or smth

inc/camera.h
u413-284-si: If we only accept normalised direction vectors, then yes. But if we do accept not normalised ones which could be a viable option as it is easier to give a direction with (1,0, 1) than calculating the normalised version of this and having to type in sth like (0.45345, 0, 0.43545). I know this isnt the normalised vector of the other btw. In that case we should put in the effort and retrieve the length of the direction  vector.

Ok, so the viewport lies at (0,0,0). That is a feasible option. And the scene may be from range_z(1, -INFINITY). 

How would your evolved version look like?

src/camera.c
u413-284-si: See above

src/camera.c
u413-284-si: Yes naming. The calculations can be discussed in person if the writing isn't clear enough.

inc/camera.h
gwolf-011235: > In that case we should put in the effort and retrieve the length of the direction vector.

If I understood you correctly we normalize the direction vector while parsing (exactly for the case that somebody puts in a not normalized one). This means the length is always 1.
I think right now (in our version) the focal length is just an arbitrary number, we don't really need it. Try to set it to 42 - it should produce the same image.

No, the viewport lies always 1 (or 42) units away from the camera. If the camera moves, so does the viewport.

Don't hassle around with a look_at point and just use the given direction vector.

src/camera.c
gwolf-011235: But we are on the same page that your suggested version produces the same result in a more roundabout way?

inc/camera.h
gwolf-011235: I did some wrong measurements, it does produce intuitive results (both versions of calculations that is).

This scratch-a-pixel article cleared up some confusion for me: [LINK](https://www.scratchapixel.com/lessons/mathematics-physics-for-computer-graphics/lookat-function/framing-lookat-function.html)

New version: 
_The camera is looking down along the negative z-axis, which is the opposite of the view direction (w)._

inc/camera.h
u413-284-si: No, the camera direction vector is not normalised when parsed. Not yet at least.

What is normalised are the normal vectors/axis of the hittable objects as this has direct influence on the hit calculations. 

Yes, so the viewport is moving with the camera. Still don't see the issue.

src/camera.c
u413-284-si: It seems that we get the same results as far as I can tell from a distance.

u413-284-si (approved):

PR #24, opened 2024-01-06
17 menu

### LINKED ISSUE #17: menu
be creative

gimme usage feedback

gwolf-011235:
#### Overlay

- The menu is an overlay, which has the same window height but a smaller width. 
- It is created while initializing mlx as extra image, to increase performance -> it can just be put over the rendered image
- The menu color, font color and width are configurable in miniRT_config
- The overlay is created by alpha blending the image with the menu color, which creates a transparent effect
- Every time the scene changes, the overlay needs to be recalculated. If only menu text changes it stays the same

#### Menu text

- The text of the menu depends on the current control mode (scene, light, camera)
- It shows info about the current hittable, light or camera
- Can also switch to show specific control schemes depending on mode.
- If menu is off (standard) displays a string with the toggle key 

#### Extend control scheme

- Lights can now be manipulated
- Menu can be toggled

#### Check for scene change

- If nothing has changed, the mlx loop exits early. If the something changed, the image gets re-rendered
- If only menu text changes only the text is updated 

#### BUG: Printing

- Printing is right now bugged. It should only work on key release, but does not
- The problem is that key events are still registered although it is printing. This means several print events can be started or even scene changes in the background.
- It seems it's not possible to clear events. If this happens to be true maybe get rid of printing.

Closes #17


inc/render.h
```diff
@@ -76,12 +79,20 @@ typedef struct s_mlx_ptrs
...
 
+typedef enum e_mode
+{
+	CTRL_SCENE,
+	CTRL_CAM,
+	CTRL_LIGHT
+}	t_mode;
+
 /**
  * @brief Overarching render struct.
```
u413-284-si, line 104: doxy needs update

inc/render.h
```diff
@@ -175,15 +206,39 @@ t_err	ft_output_as_ppm(int *img_arr, int width, int height);
...
+ * @return int		0 if successful, -1 if not.
+ */
+int		ft_keyhook_release(int key, t_render *render);
+
+// render_keyhook_scene.c
+
+/**
+ * @brief Manipulates the active hittable.
+ *
+ * Chnages the active hittable or manipulates it.
```
u413-284-si: typo

src/main.c
```diff
@@ -29,6 +29,7 @@ int	main(int argc, char **argv)
 		return (1);
 	}
 	ft_cam_init(&render.cam, render.mlx_ptrs.img.size);
+	ft_init_menu(&render.menu, MENU_OPACITY, MENU_BACK, MENU_FONT);
```
u413-284-si: Would call it ft_menu_init() or rename ft_init_cam()

src/menu_init.c
```diff
@@ -0,0 +1,23 @@
...
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "render_menu.h"
+
+void	ft_init_menu(t_menu *menu, uint8_t alpha, uint32_t colour,
+			uint32_t font_col)
+{
+	menu->inv_alpha = 255 - alpha;
+	menu->rb = (colour & 0xFF00FF) * alpha;
```
u413-284-si, line 19: How come we don't differentiate between red and blue? I read that you did for performance reasons, but why is that ok to handle them together? And why these two and not green and blue for example?

src/menu_put.c
```diff
@@ -0,0 +1,42 @@
...
+
+void	ft_menu_put_text(t_render *render)
+{
+	t_putinfo	put;
+
+	put.mlx_ptrs = render->mlx_ptrs;
+	put.pos = (t_vec2i){20, 20};
+	put.col = render->menu.font_col;
+	if (render->show_menu == false)
+		return (ft_put_str(put,
```
u413-284-si: Can that not be in one line?

src/menu_put.c
```diff
@@ -0,0 +1,42 @@
...
+				"Press I to show menu"));
+	mlx_put_image_to_window(render->mlx_ptrs.mlx_ptr,
+		render->mlx_ptrs.win_ptr, render->mlx_ptrs.veil.ptr, 0, 0);
+	put.pos = ft_put_mode(put, render->mode);
+	if (render->menu.cur_page % 3 == 0)
```
u413-284-si, line 27: Do we have MACROS for the page numbers? So instead of comparing to a number we have a description which is more readable

src/menu_put_cam_light_ctrl.c
```diff
@@ -0,0 +1,50 @@
...
+	put.pos = ft_put_ctrl_rot(put, "view");
+	put.pos = ft_put_ctrl_inc_dec(put, "FOV", true);
+	put.pos.y += Y_NEXT_LINE_BIG;
+	ft_put_str(put, "Rotate view");
+	put.pos.y += Y_NEXT_LINE;
+	ft_put_str(put, "(Mode independent)");
+	put.pos.y += Y_NEXT_LINE;
+	ft_put_str(put, "  Right mouse and move");
+	put.pos.y += Y_NEXT_LINE_BIG;
+
```
u413-284-si: Empty line 27

src/menu_put_utils.c
```diff
@@ -0,0 +1,101 @@
...
+t_vec2i	ft_put_3d_point(t_putinfo put, t_vec3 point)
+{
+	t_numinfo	numinfo;
+	const int	num_x = put.pos.x + 30;
+
+	numinfo.pad = 5;
+	numinfo.prec = 2;
+	ft_put_str(put, " x:");
+	numinfo.numf = point.x;
+	ft_put_float((t_putinfo){.mlx_ptrs = put.mlx_ptrs,
```
u413-284-si: The following lines are very similar except for the coordinates. Could be a separate function which is called 3x with the individual coordinates.

src/menu_put_utils.c
```diff
@@ -0,0 +1,101 @@
...
+
+t_vec2i	ft_put_colour(t_putinfo put, t_colour colour)
+{
+	t_numinfo	numinfo;
+	const int	num_x = put.pos.x + 70;
+
+	numinfo.pad = 3;
+	ft_put_str(put, " r:");
+	numinfo.numi = colour.r * 255.99;
+	ft_put_int((t_putinfo){.mlx_ptrs = put.mlx_ptrs,
```
u413-284-si: Could also be modularised

src/menu_put_utils.c
```diff
@@ -0,0 +1,101 @@
...
+	put.pos.y += Y_NEXT_LINE;
+	ft_put_str(put, " b:");
+	numinfo.numi = colour.b * 255.99;
+	ft_put_int((t_putinfo){.mlx_ptrs = put.mlx_ptrs,
+		.pos.x = num_x, .pos.y = put.pos.y, .col = put.col}, numinfo);
+	put.pos.y += Y_NEXT_LINE_BIG;
+	return (put.pos);
+}
+
+t_vec2i	ft_put_float_value(t_putinfo put, float value)
```
u413-284-si: Here you have a function that could be used in above cases.

src/menu_put_utils.c
```diff
@@ -0,0 +1,101 @@
...
+t_vec2i	ft_put_unit_vec(t_putinfo put, t_vec3 vec)
+{
+	t_numinfo	numinfo;
+	const int	num_x = put.pos.x + 30;
+
+	numinfo.pad = 2;
+	numinfo.prec = 5;
+	ft_put_str(put, " x:");
+	numinfo.numf = vec.x;
+	ft_put_float((t_putinfo){.mlx_ptrs = put.mlx_ptrs,
```
u413-284-si: As before.

src/parse_entity_sp_pl_cy.c
```diff
@@ -16,7 +16,7 @@ void	ft_parse_sphere(char *line, t_hittable *sphere, size_t id)
 {
 	t_sphere	*params;
 
-	sphere->id = id;
+	sphere->id = id + 1;
```
u413-284-si, line 19: Plus one due to the lights also having an ID now?

src/render_draw.c
```diff
@@ -6,12 +6,49 @@
...
 /*                                                +#+#+#+#+#+   +#+           */
 /*   Created: 2023/11/20 15:52:55 by gwolf             #+#    #+#             */
-/*   Updated: 2024/01/01 18:55:50 by gwolf            ###   ########.fr       */
+/*   Updated: 2024/01/06 12:49:14 by gwolf            ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
 #include "render.h"
 
+static uint32_t	fast_alpha_blend(uint32_t bg_color, t_menu menu)
```
u413-284-si: Our functions all have a prefix ft_ :D

src/render_keyhook_camera.c
```diff
@@ -26,17 +26,20 @@ void	ft_keyhook_rot_cam(int key, t_vec3 *direction)
 
 void	ft_keyhook_move_cam(int key, t_cam *cam)
 {
-	if (key == XK_q)
+	t_interval	lim;
+
+	lim = (t_interval){.min = FLOAT_MIN, .max = FLOAT_MAX};
```
u413-284-si: Not sure, but maybe we are allowed to do all this with declaration in one line? Need to check with norminette

src/render_keyhook_light.c
```diff
@@ -0,0 +1,59 @@
...
+
+void	ft_change_active_light(int key, t_entities *scene, int *active)
+{
+	if (*active == -1)
+		return ;
+	if (key == XK_n)
+	{
+		*active -= 1;
+		if (*active < 0)
+			*active = scene->lsrc_count + 1 - 1;
```
u413-284-si: That operation cancels out itself.

src/render_output_ppm.c
```diff
@@ -37,21 +34,47 @@ static void	ft_write_ppm_header(int width, int height, int fd)
...
-	if (ft_err_open("outfile.ppm", O_CREAT | O_WRONLY | O_TRUNC, &fd))
-		return (ERROR);
-	ft_write_ppm_header(width, height, fd);
-	pixel_sum = width * height;
 	i = 0;
 	while (i < pixel_sum)
-		ft_write_pixel(img_arr[i++], fd);
+	{
+		ft_write_pixel(pix_arr[i++], fd);
+		if (i % 100 == 0)
```
u413-284-si, line 45: What's that condition do?

u413-284-si (changes requested):
Good stuff, not many code issues discovered. As most of the results are visual, a final check shall be performed once with the rendered image to discuss design and positioning questions.

Furthermore shall talk about the mentioned printing bugs as it is not evident from just reading the source code. An example to illustrate would help.

inc/render.h
gwolf-011235: Doxy added!

inc/render.h
gwolf-011235: Chnaged.. err changed!

src/main.c
gwolf-011235: Renamed to ft_menu_init()

src/menu_put.c
gwolf-011235: You're right, leftover after refactor. Changed!

src/menu_put.c
gwolf-011235: Yes there is an enum for the pages.  The thing is that mod the third page would make less sense.
The pages are 0 scene, 1 light, 2 cam, 3 scene control, 4, light control, 5 cam control.
With mod3 I can differentiate between scene, light or cam (control or not).
I could make an extra macro NUM_OF_MODES or something to mask the three?

src/menu_put_cam_light_ctrl.c
gwolf-011235: Deleted

src/render_draw.c
gwolf-011235: Added prefix!

src/render_keyhook_camera.c
gwolf-011235: True! I made it const and norminette is accepting of it. Also changed the other one in keyhook_utils

src/render_keyhook_light.c
gwolf-011235: Yes. Maybe it was late. Deleted it :D

src/menu_init.c
gwolf-011235: It actually makes use of two simplifications:
- less variables
- less precision while division

A color is stored in a 4-byte int, where the four bytes represent alpha, red, green and blue or `ARGB`. For simple alpha blending the formula is 
ColorC = alpha{ColorA} * ColorA + (1 - alpha{ColorA} ) * ColorB. 
So we mult first color with alpha and second color with inverse of alpha. Then we add them together.

We need to multiply channel-wise, or we get funky results. This is often done by splitting the color in three variables, one for each channel (R,G,B), multiply with alpha and then stick together.
Here comes the first optimization: we only use two variables. 
Since we store the color in 4 bytes, we can make use of the remaining space. We isolate the R and B channels so `ARGB`  becomes `0R0B`. We can envision it as having two "zero points", from which we can multiply. This means R and B do not overlap, they don't interfere with each other while multiplying. For that we remove the G channel with `& 0xFF00FF`. The int now only stores R and B.  
We still need to handle G so we remove the others with `& 0x00FF00` which leaves us `00G0`.
Then we can safely multiply the channels with the inverse of alpha.
This is what happens here (the equivalent for the menu channels RB and G happens at menu_init):
```
bg_rb = (bg_color & 0xFF00FF) * menu.inv_alpha;
bg_g = (bg_color & 0x00FF00) * menu.inv_alpha;
```

We add the RB and G, each with its alpha value together. We save alpha in one byte, so it ranges from 0 - 255. This means we need to divide the resulting number by 255 (or 255.99 or whatever approximation is used) to get the actual channel value. (Alternative would be to save alpha as float in range 0.0 to 1.0 which uses 4 bytes). 
Here comes the second optimization: we bit shift 8 to the right (`>> 8`), which effectively divides by 2^8 which is 256. It won't give us the truest color blend, but a good enough approximation with speed gains. To ensure we're only left with the desired channels we `&` them again.
```
blend_rb = (((menu.rb + bg_rb) >> 8) & 0xFF00FF);
blend_g = (((menu.g + bg_g) >> 8) & 0x00FF00);
```

After that we only need to stick the two variables together which we can do with bitwise OR: `0R0B` | `00G0` = `0RGB` (we could encode the alpha in here, but we don't need it, so it is left out). This happens in the return.
`return (blend_rb | blend_g);`

src/render_output_ppm.c
gwolf-011235: While printing we output a percentage on the terminal. I added it to indicate how much longer you have to wait. Printing the image is surprisingly slow! But yeah it needs to write every single pixel to the file. 

This condition says, that we only want to output the current % for every 100 pixel. 
I think printing % for every pixel would slow it down further.

src/parse_entity_sp_pl_cy.c
gwolf-011235: No it's for the menu. In the menu the ID of the hittable is shown. It starts counting at 1, since it is more human like (three objects: 1,2,3)
The ID is (was?) never used anywhere else, so I thought I update it right here, and don't add 1 every time it gets put to screen.
The lights also start with 1, whereas ambient light has ID 0 (since its special as its not a moveable light).

Could also be reset, so that the first object has ID zero.

src/menu_put_utils.c
gwolf-011235: Good point - Refactored:
- new ft_put_coord() outputs a single coordinate
- ft_put_3d_point(), ft_put_unit_vec() and ft_put_colour() use this ft
- they also get passed a name to print (e.g. Centre or Normal)

src/menu_put_utils.c
gwolf-011235: Same as above

src/menu_put_utils.c
gwolf-011235: Not quite. This one is used for a single float value (e.g. Diameter or Height). 
It has no label on the same line like 
```
x:    50.00
``` 

The value is on the next line after the name:
```
Diameter
     50.00
```
But as mentioned above implemented your advice in other ft.
This one still was adapted - it also gets the name passed directly and now is called ft_put_single_float_value()

src/menu_put_utils.c
gwolf-011235: As before.

u413-284-si (approved):

Issue #9, opened 2023-11-03
Bonus


u413-284-si:
- intersection behaviour/hit record for 
   * cone/hyperboloid/paraboloid
- material differentiation:
   *  diffuse (recursive = Kajiya Style with defined depth, Lambert distribution)
   *  mirror/metallic
- determination of surface side
- lighting:
   * specular lighting -> Phong model
   * colored + multi-spot lights
 - colour disruption/checkerboard
 - bump map textures
 - anti-aliasing

PR #29, opened 2024-01-13
28 bonus time

### LINKED ISSUE #28: Bonus: Time
To accurately measure how long a render takes (FPS), implementation of time function is obvious.

The function is not allowed in base, but can be implemented for bonus part.

gwolf-011235:
Is set to be merged after big bonus #9 has been merged.
Closes #28 

## Overview

1. new **timestamp** function for render time 
2. add **increment** value for manipulation of scene
3. colour **interpolation**
4. overhaul of **Makefile** 
5. refactor of render options (several bools) into **bit field**.
6. refactor of **menu text**

### 1. Timestamp 

- Using the function clock_gettime() we can generate a current timestamp
- Since only allowed in bonus, the function ft_draw_scene() has two version in two files
- Base version changes nothing
- Bonus version creates a timestamp on enter. If scene was changed, redraw scene. Gen new timestamp and subtract start time. This time is saved in render 
- In bonus the time is printed in the menu as ms or fps (toggle with key 'o')
- Everything is separated in base/bonus with one caveat: last_render_time is saved in struct, but it should only be in bonus. Render has gotten far to big and I will refactor. This will split of all manipulations of scene. Then I will make a bonus version of render.h

### 2. Increment

- Manipulation of the scene depends on float increment.
- Increment can be set to low (0.1), mid (0.5) and high (1.0)
- Increment can be switched with key 'u'
- Menu shows current increment
- Increment coupled with timestamp was abandoned: since we only re-render when something changed, the longer one would wait the more movement it would make. 

### 3. Colour interpolation

- Colours ob objects / lights are now stored in colour array, in rainbow sequence
- Colours can be set directly via Numbers 1-8
- Colour can also be interpolated between the next colour of the array with 9 and 0

### 4. Makefile

- Makefile differentiates between common, base and bonus sources to specifically compile for base or bonus
- If target bonus is called make appends "_bonus" to miniRT
- targets leak, address and speed have been moved to variable MODE to set the mode independent of base or bonus
- target profile has been adapted to work with perf, which needs no recompilation
- new target help 

### 5. Bit field

- Combined different bools in t_render into one unsigned int which works as a bit field: if a bit is on the option is on and vice versa
- render now only needs one unsigned int to store several options
- Options are defined in header file render_options and render_options_bonus

### 6. Menu text
- Menu text now has two separators between header and footer.
- Small refactoring of some functions which put text
- Slight adaption of what to show when control


inc/colour.h
```diff
@@ -33,24 +33,8 @@ typedef struct s_colour
 	float	b;
 }	t_colour;
 
-typedef enum e_col_name
```
u413-284-si, line 43: If e_col_name has been moved to a different file, this has to be done in the bonus file "colour_bonus.h" as well. Sames goes for any other changes as Macros, function declarations, etc.

inc/menu_put.h
```diff
@@ -18,9 +18,22 @@
 # include "miniRT_config.h"
 # include "render.h"
 # include "vec2.h"
+# ifdef IS_BONUS
```
u413-284-si: Is implemented as 
# if IS_BONUS==1 in the other files

inc/menu_put.h
```diff
@@ -18,9 +18,22 @@
...
+# else
+#  include "render_options.h"
+# endif
 
 /* ====== MACROS ====== */
 
+# define X_MENU_OFFSET 20
+# define Y_MENU_TOP 20
+# define Y_MENU_BOTTOM_SMALL 50
+# ifdef IS_BONUS
```
u413-284-si: Same here

inc/time_bonus.h
```diff
@@ -0,0 +1,26 @@
...
+
+/* ====== LIBRARIES ====== */
+
+# include <stdint.h>
+# include <unistd.h>
+# include <time.h>
+
+/* ====== FUNCTIONS ====== */
+
+uint64_t	ft_get_time_ms(void);
```
u413-284-si, line 30: Missing doxygen

src/menu_put_text.c
```diff
@@ -0,0 +1,67 @@
...
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "menu_put.h"
+
+static t_vec2i	ft_put_header(t_putinfo put, uint32_t options)
+{
+	put.pos = ft_put_mode(put, options);
+	if (!ft_option_isset(options, O_SHOW_CTRL))
+	{
```
u413-284-si: Why the brackets

src/menu_put_time_bonus.c
```diff
@@ -0,0 +1,34 @@
...
+/*   Created: 2024/01/08 10:46:17 by gwolf             #+#    #+#             */
+/*   Updated: 2024/01/13 10:57:32 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "menu_put.h"
+
+t_vec2i	ft_put_time(t_putinfo put, uint64_t last_render_time, bool show_fps)
+{
+	const float	fps = 1000.0 / (last_render_time);
```
u413-284-si, line 17: How come we divide 1k by the time?

src/parse_entity_ACL.c
```diff
@@ -6,14 +6,15 @@
...
-/*   Updated: 2024/01/03 17:44:16 by gwolf            ###   ########.fr       */
+/*   Updated: 2024/01/13 11:00:02 by gwolf            ###   ########.fr       */
 /*                                                                            */
 /* ************************************************************************** */
 
 #include "parse.h"
 
 void	ft_parse_ambient(char *line, t_light *ambient)
 {
+	ambient->id = 0;
```
u413-284-si, line 17: Did we have an uninitialised variable before o.O

src/render_bit_field.c
```diff
@@ -0,0 +1,41 @@
...
+/*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2024/01/08 13:05:49 by gwolf             #+#    #+#             */
+/*   Updated: 2024/01/12 19:05:44 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "render_bit_field.h"
+
+bool	ft_option_isset(uint32_t bit_field, t_options option)
```
u413-284-si, line 15: Did we find a fancy video on bit operations

src/render_keyhook_colour.c
```diff
@@ -0,0 +1,82 @@
...
+	[BLACK] = (t_colour){0.0, 0.0, 0.0}};
+
+	return (colour_array[col_name]);
+}
+
+void	ft_interpolate_colour(t_colour *result, float progress)
+{
+	const int		index = (int)(progress * (NUM_RAINBOW_COLOURS - 1));
+	const float		frac = fmod(progress * (NUM_RAINBOW_COLOURS - 1), 1.0);
+	const t_colour	color1 = ft_get_colour(index);
```
u413-284-si: color and colour used in the same line xD

u413-284-si (changes requested):
Curious to see the outcome of the menu changes. No bigger issues, one major point is not to forget to transfer the changes in the base files into the bonus files when necessary after the merge. This should only be the case in a few instances as lots of files are newly created and are part of the common files. These should consider having the differentiation "IS_BONUS" in their respective header files for headers which have both a base and bonus version, such as colour.h/colour_bonus.h for example.

inc/colour.h
gwolf-011235: A very uncool message. We still need duplication :/

inc/menu_put.h
gwolf-011235: Can change.
Is there a reason for the equals 1?

src/menu_put_time_bonus.c
gwolf-011235: We want to get the frames per second.
Last_render_time is in millisecond. One second has 1000 ms.
If we divide a second by the time it takes for one frame we get fps

src/parse_entity_ACL.c
gwolf-011235: Didn't use it beforehand. 

Now we pass ambient and "normal" light as t_light and differentiate via id==0

src/render_bit_field.c
gwolf-011235: Yes! Needed a way to pass the bools around

src/menu_put_text.c
gwolf-011235: To make the diff of base and bonus more readable

inc/time_bonus.h
gwolf-011235: added

src/render_keyhook_colour.c
gwolf-011235: just much more used to color :O changed

inc/colour.h
u413-284-si: What must be done, must be done

inc/menu_put.h
u413-284-si: Tested it without == 1, works as well. Can be omitted.

But there is a difference between #if and #ifdef. The first checks if the following statement is true, the latter checks if the entity has been previously defined. It could be that a macro has been defined with # define STH, but STH = false. Then the conditionals would give different results. Better and safer to use the first.

src/menu_put_text.c
u413-284-si: Unnecessary lines of code (syntax) imo

src/render_bit_field.c
u413-284-si: and a bool array would have been to simple xD

inc/colour.h
gwolf-011235: True.

inc/menu_put.h
gwolf-011235: Didn't know that! Changed to if - also in bit_field

src/menu_put_text.c
gwolf-011235: Ok removed

src/render_bit_field.c
gwolf-011235: It looks more than it is really. The bit shifting part is really only flipping a specific bit on or off. No fancy double flips or something. The functions make it easy to interact. 
Bool array would be bigger after 4 bools. Makes it easy to add in more options (as long as not all 32 bits have been used).

I like.

u413-284-si: Yes I can tell you like. Me gusta tambien pero es fancy.

u413-284-si (approved):

Issue #20, opened 2023-12-24
BUG: Strong light intensity causes inversion of colours


u413-284-si:
Light source intesities > 0.9 lead to inversion of white/near white colours and create instead black patches where it should be white.

Issue #15, opened 2023-11-21
Valgrind: uninitialised values


gwolf-011235:
The function ft_render_image generates a lot of errors (608658) in valgrind. This makes it hard to check, if new errors are introduced.

I suspect it has to do with t_hitrecord, but i'm not sure what do initialise it to.

Maybe there are also other errors.

PR #30, opened 2024-01-15
9 bonus

### LINKED ISSUE #9: Bonus

u413-284-si:
Implements various additional features to the basic ray tracing program:

1) Intersection with a cone
+ file checking factoring in cone line
+ entity count amended with cone
+ parsing of cone parameters
+ hit management of cone including mathematical description of cone geometries

2) Different material behaviour
+ diffuse material
+ reflective material (recursive ray calculation, amount of recursions=reflections configurable)

3) Lighting
+ additional specular lighting -> leading to full Phong modell
+ multiple light sources > 1
+ coloured lights

3) Textures
+ introduction of UV-coordinates to map 3D geometries to 2D plane
+ checkered texture on all hittable geometries enabled via mathematical translation in UV-coordinates

4) Anti-Aliasing
+ optimisation of edge rendering at high contrast transitions to reduce "ruggedness"
+ creation of randomised sample rays around target pixel
+ amount of sample rays configurable

Additionally fixes following bugs:

+ Sphere parsing and rendering ( diameter instead of radius is parsed -> following calculations adapted)
+ Cylinder geometry calculation and rendering
+ Plane rendering
+ Colour overflow while calculating lighting influence on ray colour leading to colour inversion
+ Uninitialised values for hit record, colours and various other variables

Closes #9 , #10 , #15 , #20


Makefile
```diff
@@ -21,6 +21,17 @@ else
...
+# *     Verbosity              *
+# ******************************
+
+# Set VERBOSE=1 to echo all commands
+ifeq ($(VERBOSE),1)
+	SILENT =
+else
+	SILENT = @
+endif
+
```
gwolf-011235: Double. Leftover from merge?

Makefile
```diff
@@ -82,118 +106,187 @@ LDLIBS := -lft -lm -lmlx -lXext -lX11
 CC := cc
 CPPFLAGS := -I $(INC_DIR) -I lib/libft/include
 CFLAGS = -Wall -Werror -Wextra -gdwarf-4
+CFLAGS = -Wall -Werror -Wextra -gdwarf-4
```
gwolf-011235: Double. Leftover from merge?

Makefile
gwolf-011235, line 1: Makefile has  variables, which are not needed anymore. I think it makes more sense to clean up in other branch

inc/camera_bonus.h
```diff
@@ -0,0 +1,188 @@
...
+
+#ifndef CAMERA_BONUS_H
+# define CAMERA_BONUS_H
+
+/* ====== LIBRARIES ====== */
+
+# include "utils_bonus.h"
+# include "miniRT_config_bonus.h"
+# include "colour_bonus.h"
+# include "entities_bonus.h"
```
gwolf-011235: Not needed includes.
Also utils_bonus.h not needed
vec3.h is missing (though still compiles - include from somewhere else)

inc/camera.h
gwolf-011235, line 1: camera.h and camera_bonus.h are very similar > diff shows includes and two functions.
camera_bonus.h includes camera.h and adds its functions.

inc/camera_bonus.h
```diff
@@ -0,0 +1,188 @@
...
+ * The pixel colour is generated as an average of a sample of ray colours
+ * around the pixel in question.
+ * @param iterate		Integer array containing current pixel position
+ * @param pixels		Pixel grid defining its starting position and
+ * 						horizontal/vertical distances
+ * @param cam			Camera
+ * @param scene			The scene containing all hittables
+ * @return t_colour
+ */
+t_colour	ft_anti_aliase_colour(int iterate[2], t_pixel_grid pixels, \
```
gwolf-011235: I don't think this is the right file for these functions (pixel_sample and anti_). 
It's more like a render function: take ray direction and randomly move it.

src/camera_bonus.c
gwolf-011235, line 1: File is identical to base except for last function > move out function.
As said above - doesn't fit that good into cam IMO

src/camera_bonus.c
```diff
@@ -0,0 +1,81 @@
...
+	cam->pixels.delta_v = ft_vec3_scale(viewport_v, (1.0 / cam->image.y));
+	viewport_upper_left = ft_vec3_sub(ft_vec3_sub(ft_vec3_sub(\
+		cam->centre, ft_vec3_scale(cam->w, cam->focal_length)), \
+		ft_vec3_scale(viewport_u, 0.5)), \
+		ft_vec3_scale(viewport_v, 0.5));
+	cam->pixels.pos00 = ft_vec3_add(viewport_upper_left, ft_vec3_scale(\
+		ft_vec3_add(cam->pixels.delta_u, cam->pixels.delta_v), 0.5));
+}
+
+t_colour	ft_anti_aliase_colour(int iterate[2], t_pixel_grid pixels, \
```
gwolf-011235: Use t_vec2i struct instead of iterate array

src/check_entity_sp_pl_cy_bonus.c
```diff
@@ -0,0 +1,117 @@
...
+	ft_rm_space(&line);
+	if (!ft_isvalid_float_block(&line, -1.0, 1.0))
+		return (UNKNOWN);
+	i = -1;
+	while (++i < 5)
+	{
+		ft_rm_space(&line);
+		if (!ft_isvalid_float(&line, FLOAT_MIN, FLOAT_MAX, false))
+			return (UNKNOWN);
+	}
```
gwolf-011235, line 80: Looks like a norm fix.
Could also be implemented in sphere and plane for consistency

src/colour.c
```diff
@@ -35,37 +35,17 @@ t_colour	ft_add_colour(t_colour c1, t_colour c2)
 
 t_colour	ft_hadamard_colour(t_colour c1, t_colour c2)
 {
+	t_interval	interval;
+
+	interval.min = 0.0;
+	interval.max = 1.0;
```
gwolf-011235: Could be made const and init with compound literal
const t_interval interval = (t_interval){.min = 0, .max = 0}

src/colour_bonus.c
```diff
@@ -0,0 +1,90 @@
...
+	interval.max = 0.999;
+	interval.min = 0.0;
+	pixel.r *= scale;
+	pixel.g *= scale;
+	pixel.b *= scale;
+	ir = (int)(256 * ft_clamp(pixel.r, interval));
+	ig = (int)(256 * ft_clamp(pixel.g, interval));
+	ib = (int)(256 * ft_clamp(pixel.b, interval));
+	printf("%d %d %d\n", ir, ig, ib);
+}
```
gwolf-011235: We don't need this anymore.
It was for printing the file at first without mlx. Lives on as slightly different version in output_ppm
It's not used anywhere right now

src/colour.c
```diff
@@ -35,37 +35,17 @@ t_colour	ft_add_colour(t_colour c1, t_colour c2)
...
 	return ((t_colour){
-		.r = c1.r * c2.r,
-		.b = c1.b * c2.b,
-		.g = c1.g * c2.g
+		.r = ft_clamp(c1.r * c2.r, interval),
+		.b = ft_clamp(c1.b * c2.b, interval),
+		.g = ft_clamp(c1.g * c2.g, interval)
```
gwolf-011235, line 31: Is clamping everytime expected? Colour add for example does not clamp automatically

src/hit.c
```diff
@@ -6,21 +6,30 @@
...
-	t_interval ray_d)
+void	ft_init_hitrecord(t_hitrecord *rec)
+{
+	rec->axis_hit = (t_vec3){.x = 0, .y = 0, .z = 0};
+	rec->colour = (t_colour){.b = 0, .g = 0, .r = 0};
+	rec->d = INFINITY;
+	rec->normal = (t_vec3){.x = 0, .y = 0, .z = 0};
+	rec->point = (t_vec3){.x = 0, .y = 0, .z = 0};
+	rec->ray = (t_ray){.origin = {0}, .direction = {0}, .d = 0};
+}
```
gwolf-011235, line 23: Why not use your favorite: compound literal?
rec = (t_hitrecord){ .d = INFINITY };
does the same as this function.

src/hit_cone_bonus.c
```diff
@@ -0,0 +1,105 @@
...
+	potential_hits[1] = eq.d2;
+	potential_hits[2] = ft_co_cap_hit(co, ray);
+	return (true);
+}
+
+float	ft_co_cap_hit(t_cone co, t_ray ray)
+{
+	t_plane		pl;
+	t_hitrecord	rec;
+	t_interval	ray_d;
```
gwolf-011235: Can be made const and assigned

src/hit_cone_bonus.c
```diff
@@ -0,0 +1,105 @@
...
+
+	pl.point = co.base;
+	pl.normal = co.axis;
+	pl.colour = co.colour;
```
gwolf-011235: Could be assigned at the top to const t_plane

src/hit_cylinder_bonus.c
```diff
@@ -0,0 +1,108 @@
...
+	t_equation	eq;
+	t_vec3		cap1_ray;
+
+	cap1_ray = ft_vec3_sub(ray.origin, cy.cap1);
+	eq.a = ft_vec3_dot(ray.direction, ray.direction) - \
+		pow(ft_vec3_dot(ray.direction, cy.axis), 2);
+	eq.b = 2.0 * (ft_vec3_dot(ray.direction, cap1_ray) - \
+		(ft_vec3_dot(ray.direction, cy.axis) * ft_vec3_dot(cap1_ray, cy.axis)));
+	eq.c = ft_vec3_dot(cap1_ray, cap1_ray) - \
+		pow(ft_vec3_dot(cap1_ray, cy.axis), 2) - pow(cy.d / 2.0, 2);
```
gwolf-011235: ft_vec3_dot(ray.direction, cy.axis) and ft_vec3_dot(cap1_ray, cy.axis) are calculated twice. 
Would fit in a variable

src/hit_cylinder_bonus.c
```diff
@@ -0,0 +1,108 @@
...
+		rec->d = potential_hits[1];
+	if (ft_cy_check_cap(cy, cy.cap1, potential_hits[2], rec))
+		rec->d = potential_hits[2];
+	if (ft_cy_check_cap(cy, cy.cap2, potential_hits[3], rec))
+		rec->d = potential_hits[3];
+	if (rec->d == INFINITY)
+		rec->d = 0;
+}
+
+void	ft_cy_set_hitrecord(t_cylinder cy, t_hitrecord *rec, t_hitrecord tmp)
```
gwolf-011235, line 95: Similar function could be made for cone for consistency.

src/hit_plane.c
```diff
@@ -3,33 +3,35 @@
...
 	t_equation	eq;
 
-	eq.b = ft_vec3_dot(ray.direction, pl.normal);
-	if (eq.b < EPSILON)
-		return (false);
-	pl_point_ray = ft_vec3_sub(ray.origin, pl.point);
 	eq.a = 0.0;
-	eq.c = ft_vec3_dot(pl_point_ray, pl.normal);
-	if (ft_solve(&eq) < 0)
+	eq.b = ft_vec3_dot(ft_vec3_norm(rec->ray.direction), pl.normal);
```
gwolf-011235: The ray direction should already be normalized here right?

src/hit_plane_bonus.c
```diff
@@ -0,0 +1,51 @@
...
+	ft_get_pl_uvcoords(rec);
+	return (true);
+}
+
+void	ft_get_pl_uvcoords(t_hitrecord *rec)
+{
+	t_vec3	base1;
+	t_vec3	base2;
+
+	base1 = ft_vec3_norm(ft_vec3_cross(rec->normal, (t_vec3){1, 0, 0}));
```
gwolf-011235: rec->normal is plane_normal, which should be normalized. You cross it with a normal vector which gives a normal vector. ft_vec3_norm is therefore not needed

src/hit_plane_bonus.c
```diff
@@ -0,0 +1,51 @@
...
+	if (ft_vec3_equal(base1, (t_vec3){0, 0, 0}))
+		base1 = ft_vec3_norm(ft_vec3_cross(rec->normal, (t_vec3){0, 0, 1}));
+	base2 = ft_vec3_norm(ft_vec3_cross(rec->normal, base1));
```
gwolf-011235: Same here

src/hit_sphere.c
```diff
@@ -3,26 +3,25 @@
...
 	float		closest_d;
 
-	sp_centre_ray = ft_vec3_sub(ray.origin, sp.centre);
-	eq.a = ft_vec3_dot(ray.direction, ray.direction);
-	eq.b = 2.0 * ft_vec3_dot(ray.direction, sp_centre_ray);
-	eq.c = ft_vec3_dot(sp_centre_ray, sp_centre_ray) - pow(sp.r, 2);
+	sp_centre_ray = ft_vec3_sub(rec->ray.origin, sp.centre);
+	eq.a = ft_vec3_dot(rec->ray.direction, rec->ray.direction);
+	eq.b = 2.0 * ft_vec3_dot(rec->ray.direction, sp_centre_ray);
+	eq.c = ft_vec3_dot(sp_centre_ray, sp_centre_ray) - pow(sp.d / 2.0, 2);
```
gwolf-011235: Why are we storing the diameter if we always divide it by two?
Every time this function gets called we need to divide by two. Compiler can't optimize since we don't know the value beforehand 
We could parse it as diameter and store it as radius

src/hit_sphere_bonus.c
```diff
@@ -0,0 +1,49 @@
...
+	{
+		closest_d = eq.d2;
+		if (!ft_surrounds(closest_d, ray_d))
+			return (false);
+	}
+	rec->d = closest_d;
+	rec->point = ft_ray(rec->ray, closest_d);
+	rec->normal = ft_vec3_norm(ft_vec3_scale(\
+		ft_vec3_sub(rec->point, sp.centre), (1.0 / (sp.d / 2.0))));
+	rec->colour = sp.colour;
```
gwolf-011235: Could be made into function like for hit_cylinder

src/parse_entity_sp_pl_cy.c
```diff
@@ -33,6 +33,7 @@ void	ft_parse_plane(char *line, t_hittable *plane, size_t id)
 	params = &plane->params.pl;
 	ft_parse_float_block(&line, &params->point);
 	ft_parse_float_block(&line, &params->normal);
+	params->normal = ft_vec3_norm(params->normal);
```
gwolf-011235, line 37: I think this is the only check" we have if the passed vector is a normal vector right? So if we get 0,0,0 it bad?

src/ray.c
```diff
@@ -26,14 +25,15 @@ t_colour	ft_ray_colour(t_ray ray, t_entities scene)
 	t_interval	ray_d;
 	int			i;
 
-	ray_d.min = 0;
-	ray_d.max = INFINITY;
-	rec.d = INFINITY;
+	ft_init_colour(&ray_colour);
+	ft_init_hitrecord(&rec);
+	rec.ray = ray;
+	ft_init_interval(&ray_d);
```
gwolf-011235: interval can be made const and assigned

src/ray_bonus.c
```diff
@@ -0,0 +1,68 @@
...
+/*   Updated: 2024/01/15 12:02:10 by sqiu             ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "lighting_bonus.h"
+
+t_vec3	ft_ray(t_ray ray, float d)
+{
+	return (ft_vec3_add(ray.origin, ft_vec3_scale(ray.direction, d)));
+}
```
gwolf-011235: I saw this elsewhere: would it make more sense to call it ft_scale_ray() ? 
Doxy says "Represents a ray", but you need to pass a ray.

src/ray_bonus.c
```diff
@@ -0,0 +1,68 @@
...
+
+t_colour	ft_ray_colour(t_ray ray, t_entities scene, t_cam cam)
+{
+	t_colour	ray_colour;
+	t_hitrecord	rec;
+	t_hittable	cur;
+	t_interval	ray_d;
```
gwolf-011235, line 25: Interval to const

src/ray_bonus.c
```diff
@@ -0,0 +1,68 @@
...
+	t_colour	blue;
+	t_colour	ray_colour;
+
+	white.r = 1.0;
+	white.g = 1.0;
+	white.b = 1.0;
+	blue.r = 0.5;
+	blue.g = 0.7;
+	blue.b = 1.0;
+	unit_direction = ft_vec3_norm(ray.direction);
```
gwolf-011235: For this and bonus: ft_vec3_norm() not needed - direction is normed anyways

src/render_draw.c
```diff
@@ -69,26 +69,28 @@ void	ft_put_pix_to_image(t_img *img, int x, int y, int color)
 void	ft_render_image(t_render *render)
 {
 	t_ray		ray;
-	int			y;
-	int			x;
+	int			iterate[2];
```
gwolf-011235: Use t_vec2i instead

src/render_draw_bonus.c
```diff
@@ -0,0 +1,90 @@
...
+			*pixel++ = (color >> i) & 0xFF;
+		else
+			*pixel++ = (color >> (img->bpp - 8 - i)) & 0xFF;
+		i -= 8;
+	}
+}
+
+void	ft_render_image(t_render *render)
+{
+	int			iterate[2];
```
gwolf-011235: Use t_vec2i instead

src/render_draw_bonus.c
```diff
@@ -0,0 +1,90 @@
...
+	t_colour	pixel_colour;
+	int			colour;
+
+	iterate[1] = -1;
+	while (++iterate[1] < render->cam.image.y)
+	{
+		iterate[0] = -1;
+		while (++iterate[0] < render->cam.image.x)
+		{
+			pixel_colour = ft_anti_aliase_colour(iterate, \
```
gwolf-011235, line 81: Do we always anti-aliase in bonus? Or should we be able to turn it of?

src/scene_init.c
gwolf-011235, line 1: Do we keep this file around? Since we can pretty easily init objects via parsing now

src/scene_light_bonus.c
```diff
@@ -0,0 +1,100 @@
...
+
+t_colour	ft_specular_light(t_light cur, t_hitrecord rec, t_cam cam)
+{
+	float	specular_strength;
+	float	specular_factor;
+	t_vec3	view_direction;
+	t_vec3	reflect_direction;
+	t_vec3	light_direction;
+
+	specular_strength = 0.5;
```
gwolf-011235: Is specular_strength always 0.5? The a macro would be better

src/utils_camera_bonus.c
```diff
@@ -0,0 +1,39 @@
...
+#include "ray_bonus.h"
+
+t_ray	ft_create_sample_ray(int i, int j, t_pixel_grid pixels, t_cam cam)
+{
+	t_vec3	pixel_centre;
+	t_vec3	pixel_sample;
+	t_ray	ray;
+
+	pixel_centre = ft_vec3_add(ft_vec3_add(pixels.pos00, \
+		ft_vec3_scale(pixels.delta_u, i)), ft_vec3_scale(pixels.delta_v, j));
```
gwolf-011235: This recalculates the ray direction for every sample, per pixel. 

I think we can easily optimize it if we calculate the ray directions once for the current cam position. Then for AA produce sample rays depending on these directions.
Which is why I don't like this function in this file: ray directions would be camera dependent. However AA would be more render dependent. I suggest just put these two into new file anti_aliasing

src/utils_colour.c
```diff
@@ -0,0 +1,44 @@
...
+/* ************************************************************************** */
+
+#include "colour.h"
+
+void	ft_init_colour(t_colour *colour)
+{
+	colour->b = 0;
+	colour->g = 0;
+	colour->r = 0;
+}
```
gwolf-011235: A colour could just be init with compound literal:
colour = (t_colour){ 0 };

src/utils_cylinder.c
```diff
@@ -6,72 +6,71 @@
...
 
-	hit = ft_ray(ray, rec.d);
 	if (ft_vec3_equal(rec.axis_hit, cy.cap1))
 		normal = ft_vec3_scale(cy.axis, -1);
 	else if (ft_vec3_equal(rec.axis_hit, cy.cap2))
-		normal = ft_vec3_scale(cy.axis, 1);
+		normal = cy.axis;
 	else
-		normal = ft_vec3_sub(hit, rec.axis_hit);
+		normal = ft_vec3_sub(rec.point, rec.axis_hit);
```
gwolf-011235: I think this should be normalized

src/utils_cylinder_bonus.c
```diff
@@ -0,0 +1,101 @@
...
+t_vec3	ft_cy_normal(t_hitrecord rec, t_cylinder cy)
+{
+	t_vec3	normal;
+
+	if (ft_vec3_equal(rec.axis_hit, cy.cap1))
+		normal = ft_vec3_scale(cy.axis, -1);
+	else if (ft_vec3_equal(rec.axis_hit, cy.cap2))
+		normal = cy.axis;
+	else
+		normal = ft_vec3_sub(rec.point, rec.axis_hit);
```
gwolf-011235: Normalize?

miniRT_supreme
gwolf-011235, line 1: Binary file pushed. Should be deleted. Old gitignore didn't check for miniRT* but for miniRT_speed and friends, so miniRT_supreme has been pushed (git add . ? )

src/utils_random_bonus.c
```diff
@@ -0,0 +1,23 @@
...
+/*   Created: 2023/11/19 20:20:40 by u413q             #+#    #+#             */
+/*   Updated: 2023/12/25 20:31:42 by sqiu             ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "utils_bonus.h"
+
+float	ft_random_float(void)
+{
+	return (rand() / (RAND_MAX + 1.0));
```
gwolf-011235, line 18: Since srand() is not called anywhere, we always have the same seed i.e. start with srand(1)

inc/utils_bonus.h
```diff
@@ -0,0 +1,197 @@
...
+ * @brief Quaternion definition with
+ * 
+ * @param scalar	Real component
+ * @param vector	Imaginary component consisting of three values/directions
+ */
+typedef struct s_quaternion
+{
+	float	scalar;
+	t_vec3	vector;
+}	t_quaternion;
```
gwolf-011235, line 79: This would fit pretty well into its own header quaternion.h

inc/utils_bonus.h
```diff
@@ -0,0 +1,197 @@
...
+ * @brief Solves the given equation
+ * 
+ * If the parameter a = 0, the equation is linear and the sole
+ * solution is calculated and stored in d1.
+ * Otherwise the equation is quadratic and dependent on the
+ * discriminant real solutions exist or not:
+ * 
+ * Discriminant < 0: no real solution, return -1
+ * Discriminant = 0: one solution, return 0
+ * Discriminant > 0: two solutions, return > 0
```
gwolf-011235, line 129: We could also add:
The solutions are stored in d1 and d2

inc/utils_bonus.h
```diff
@@ -0,0 +1,197 @@
...
+float			ft_random_float(void);
+
+/**
+ * @brief Return a random float within min and max
+ * 
+ * @param min 		Lower bounder
+ * @param max 		Upper bounder
+ * @return float 
+ */
+float			ft_random_float_in(float min, float max);
```
gwolf-011235: More like ft_random_float_clamp

inc/utils_bonus.h
```diff
@@ -0,0 +1,197 @@
...
+ * @param interval 			Interval to be initiated
+ */
+void			ft_init_interval(t_interval *interval);
+
+/**
+ * @brief 
+ * 
+ * @param q1 
+ * @param q2 
+ * @return t_quaternion
```
gwolf-011235: Missing

inc/utils_bonus.h
```diff
@@ -0,0 +1,197 @@
...
+t_quaternion	ft_quaternion_mult(t_quaternion q1, t_quaternion q2);
+
+/**
+ * @brief 
+ * 
+ * @param vec 
+ * @param axis 
+ * @param angle 
+ * @return t_vec3 
+ */
```
gwolf-011235: Missing

inc/ray_bonus.h
```diff
@@ -0,0 +1,81 @@
...
+ * and creates a transition between two colours
+ * @param ray 			Ray shot into scene
+ * @return t_colour 
+ */
+t_colour	ft_background_colour(t_ray ray);
+
+/**
+ * @brief Creates a random sample ray in a square surrounding the current pixel
+ * 
+ * Square size is the distance to the next pixel. Therefore it extends halfway
```
gwolf-011235: "Square side length"

inc/parse_bonus.h
```diff
@@ -0,0 +1,222 @@
...
+void	ft_parse_cylinder(char *line, t_hittable *cylinder, size_t id);
+
+/**
+ * @brief Parse cone line.
+ * 
+ * Function gets called if line has identifier "co".
+ * Parse the following:
+ * - x,y,z coordinates of the apex of the cone.
+ * - 3d normalized axis vector of cone. In range [-1,1] for each x,y,z axis.
+ * - the cone radius.
```
gwolf-011235, line 176: For unifomity we should also parse the cones diameter

gwolf-011235 (changes requested):
The result of the render looks amazing!

We still have much code duplication. For me it was often hard to spot the difference (if there was any) between base and bonus file. But also your patience is running out.
Last proposal: we add the ifdef guard to the common files (since they need to check for bonus or not). The header files don't have the guard (sometimes I'm not sure if they are even doing what we want)

Re implementation: most of the time I have no idea whats happening :D I'm not sure if we want / can extend doxygen, but often times it just repeats the function name (e.g. ft_specular_light - Create specular lighting).
So I just go by eye and say: yay!

I think there are some easy performance gains (e.g. don't normalize twice; store radius instead of diameter)

I'd be happy with movement of ifdef guards and radius instead of diameter

Makefile
u413-284-si: yup, changed

Makefile
u413-284-si: removed

Makefile
u413-284-si: agreed

inc/camera_bonus.h
u413-284-si: Colour and entities headers are required for the function:

t_colour	ft_anti_aliase_colour(int iterate[2], t_pixel_grid pixels, \
	t_cam cam, t_entities scene);

Otherwise it throws problems as it doesn't recognize both typedefs:
image

Weirdly enough it still compiles though.. would expect it to return an unknown type error. Would still keep both headers in cause this behaviour seems unreliable and not clean.

u413-284-si: Removed utils_bonus.h

inc/camera.h
u413-284-si: Yes, this is how the bonus part works. You take the base part and build upon that extending its capacities.

inc/camera_bonus.h
u413-284-si: moved to ray_bonus.h

src/camera_bonus.c
u413-284-si: moved ft_anti_aliase_colour(int iterate[2], t_pixel_grid pixels, t_cam cam, t_entities scene) into render_draw_bonus.c
moved ft_pixel_sample(t_pixel_grid pixels) and ft_create_sample_ray(int i, int j, t_pixel_grid pixels, t_cam cam); into ray_bonus.c

-> deleted camera_bonus.c, camera_bonus.h, utils_camera_bonus.c
-> camera.c is common file
-> one camera header

src/camera_bonus.c
u413-284-si: changed

src/check_entity_sp_pl_cy_bonus.c
u413-284-si: done

src/colour.c
u413-284-si: done

src/colour_bonus.c
u413-284-si: deleted, the same should be done with the whole scene_init.c file imo

src/colour.c
u413-284-si: It's only necessary in ft_hadamard_colour as it is the last operation in the lighting calculation. 

On the contrary it would falsify the result if applied to ft_add_colour as information would get lost in the middle of the process. So it should only happen at this point.

src/hit.c
u413-284-si: The other values still have to be set to 0

src/hit_cone_bonus.c
u413-284-si: also the case for the cylinder, implemented in both

src/hit_cone_bonus.c
u413-284-si: see above

src/hit_cylinder_bonus.c
u413-284-si: replaced with variables, similar enhancement effected in cone calculation

src/hit_cylinder_bonus.c
u413-284-si: There is a pendant in cone

src/hit.c
gwolf-011235: If only one element is specified, the rest is automatically init to 0

As explained in PR #18 this is guaranteed by C standard.

src/hit_cylinder_bonus.c
gwolf-011235: Sorry! Looked at the wrong file, was in hit_cylinder.c

Could the base file also have this function (even if its not needed by norm)? Would make diff easier

src/hit.c
u413-284-si: Wasn't that the case only with static variables?

gwolf-011235: No.

From #18: 
The C99 standard guarantees it:
6.7.8 paragraph 19 "[...] all subobjects that are not initialized explicitly shall be initialized implicitly the same as objects that have static storage duration."

src/hit_cylinder_bonus.c
u413-284-si: Done, cause I feel charitable today due to my bday :D

src/hit_plane.c
u413-284-si: It has been normalised in render draw. That is probably a leftover of the plane debugging. Removed in both base and bonus.

src/hit_plane_bonus.c
u413-284-si: True, removed

src/hit_plane_bonus.c
u413-284-si: removed

src/hit_sphere.c
u413-284-si: Changed parsing to radius and adapted all instances of usage to radius

src/hit_sphere_bonus.c
u413-284-si: Not really necessary

src/parse_entity_sp_pl_cy.c
u413-284-si: At this point we should not receive a null-vector. That should be blocked as wrong input at the checker.

src/ray.c
u413-284-si: No, as the interval max value needs to be adapted further on to exclude all hittables which are farther away than the current hit. 
if (rec.d < ray_d.max)
	ray_d.max = rec.d;

src/ray_bonus.c
u413-284-si: Renamed

src/ray_bonus.c
u413-284-si: see above

src/ray_bonus.c
u413-284-si: CHanged

src/render_draw.c
u413-284-si: Changed

src/render_draw_bonus.c
u413-284-si: Changed

src/render_draw_bonus.c
u413-284-si: We can set the number of samples to 1. It is not exactly the same as turning it off as the ray would still get randomly offset, but the difference would be barely visible to the human eye

src/scene_init.c
u413-284-si: That's what I mentioned before, I think we  can delete it.

src/scene_light_bonus.c
u413-284-si: This is a standard value accepted commonly. We could make it configurable, but there is no real benefit as the strength of the specular effect is already determined by the shininess

src/ray.c
gwolf-011235: Then assignment of (t_interval){.min = 0, .max = 0} would suffice.

Argument against this way of init function: the variable ray_d lies on the stack (and dies after this function call). We need to pass the address of it to this init function. This means the compiler can't really optimize, he doesn't know the address beforehand. This makes it slow.
If the init function would return an initializes t_interval it could be optimized.
This of course begs the question why even make it a function if we can just assign it compound literal it.

src/render_draw_bonus.c
gwolf-011235: I think it would be a noticeable speed difference.
It could be a bool which chooses the function.

In this context: why even have random samples and not fixed ones (up, down, left, right), if one random sample would effectively the same as other function? Would increase speed since no random calculations.

src/scene_light_bonus.c
gwolf-011235: Benefit would be readablity

src/utils_camera_bonus.c
u413-284-si: Moved pixel centre into ft_anti_aliase_colour so that it is only performed once per pixel.  The functions have already been moved to ray_bonus.c / render_draw_bonus.c, utils_camera_bonus.c is deleted.

src/utils_colour.c
u413-284-si: removed and replaced with compound literal

src/utils_cylinder.c
u413-284-si: it is normalised in ft_set_cylinder

src/utils_cylinder_bonus.c
u413-284-si: see above

miniRT_supreme
u413-284-si: yeah we can delete it, was pushed by mistake

src/utils_cylinder.c
u413-284-si: actually moved it to this point, so that is doesn't normalise the axes which are already normalised

u413-284-si: done for cone as well

src/utils_random_bonus.c
u413-284-si: added srand(time(0)) to have different numbers at each program call

inc/utils_bonus.h
u413-284-si: Do we rly need an extra header for that? All the other math functions are also under utils.h/utils_bonus.h

inc/utils_bonus.h
u413-284-si: Added

inc/utils_bonus.h
u413-284-si: There is a difference: the clamp function actually cuts down a passed float into the interval whereas this function generates a float in the interval. The clamping is more aggressive so to say.

inc/utils_bonus.h
u413-284-si: Added

inc/utils_bonus.h
gwolf-011235: No probs not

inc/utils_bonus.h
u413-284-si: Added

src/parse_entity_sp_pl_cy.c
gwolf-011235: OK will implement

inc/ray_bonus.h
u413-284-si: Changed

inc/parse_bonus.h
u413-284-si: Or we could just save the radius of the cylinder as with the sphere because we only use the radius everywhere. The cone only needs the radius and as it is a bonus we can decide what is given in the .rt file

inc/utils_bonus.h
gwolf-011235: More of a readability question. 
I think the intention is to read it as "random float in float min float max"? 
With clamp the intention is clear, disregarding whats happening behind the scene.

If not: refactor into t_interval?

src/hit.c
u413-284-si: Spicy

inc/parse_bonus.h
gwolf-011235: Yeah of course store it as radius.

It's a legacy question: we have two existing things which want a diameter. Even though that's bad design we still need to adhere to the standard (we solve this issue by just storing it as radius).

Now we add a new thing. Since the old two things want diameter it would be more uniform to ask the user for diameter of the cone. He has no idea that that's actually bad, but he would see 
- cylinder->diameter; 
- sphere->diameter; 
- cone->radius.

src/scene_light_bonus.c
u413-284-si: How so? Rather to the contrary as it introduces a value to the user interface (config file) which he probably does not understand. This way it is hidden and the user does not have to concern about it.

src/render_draw_bonus.c
u413-284-si: Cause random samples are cool. Of course it costs time. But it's cool.

I mean I can implement a simple bool but we'd had to run it with a performance tester to see if it rly does make a difference

gwolf-011235: If the difference in the final picture is negligible, imo it would be more cool if we render faster.

Right now we have to wait pretty long for one bonus render. Multithreading can speed up, but not that much. 

We can run a performance tester (perf is my choice in makefile, since it doesn't need recompile) but just simple math: 
- pix_centre - cam->centre 
- generate random_float; add to pix_centre; pix_centre - cam->centre

src/scene_light_bonus.c
gwolf-011235: I wouldn't put it in config (as we don't want to configure it) but in the header of lighting. 
Right now it's a magic number and its nowhere explained why its that way.

src/hit.c
gwolf-011235: As is every good documentation 🌶️

src/render_draw_bonus.c
gwolf-011235: Some data:
- Call ft_render_image directly from main (not entering mlx loop)
- One with base version; other with bonus (anti_alias, SAMPLE_SIZE set to 1)
- 3 runs, run with perf

Base: 4220,485 ms (average)
Anti-alias: 5561,727 ms (average)

So about 1,3 sec difference

Length of function calls is a bit misleading since different stuff happens.

As you predicted, the difference is neglible
comparison

I can add bool to switch between versions.

src/hit.c
u413-284-si: Changed into compound literal, removed init function

u413-284-si: Ok, after implementing this, the program became unstable. This implicit assignment of variables seems to be unreliable, leading to segfaults when running more demanding scenes which were not present before. After reverting back to ft_init function it is once again stable.

src/ray.c
u413-284-si: Replaced with compound literal

src/render_draw_bonus.c
u413-284-si: Good testing, quite insightful. Let's discuss together how you imagine that switch

src/scene_light_bonus.c
u413-284-si: macro added in header, replaced variable

inc/utils_bonus.h
u413-284-si: Proposal: ft_random_float_within()

inc/parse_bonus.h
u413-284-si: So you wanna put diameter for the cone in the .rt file, but store and process it behind the scenes as radius? What about your display in the menu? There it should be again diameter?

inc/utils_bonus.h
gwolf-011235: Noice

inc/parse_bonus.h
gwolf-011235: Yes, as the same thing happens for sphere and cylinder.

Display should be in diameter again. I'd argue we don't need an extra variable for  it, since the calc *2 only happens once per menu page and not per pixel

src/hit.c
gwolf-011235: OK then not

inc/utils_bonus.h
u413-284-si: Done

inc/parse_bonus.h
u413-284-si: Done

gwolf-011235 (approved):

PR #31, opened 2024-01-18
23 check for valid orientation vectors

### LINKED ISSUE #23: Check for valid orientation vectors
While checking input all given _3d normalized orientation vectors_ should be checked, if they are indeed unit vectors.

- If not: vector could be normalized anyway and just print out a warning
- Special case (0,0,0): can't be normalized - in this case abort?

gwolf-011235:
#### Adds check for unit vectors 
- ft_isvalid_unit_vec() works like ft_isvalid_float_block()
- Additionally keeps track of checksum of all three floats 
- If checksum == 0.0 throws error
- If checksum > 1.0 prints warning

Implemented for camera, plane and cylinder

#### Adds check for len in normalize 
- ft_vec3_norm() calcs len of vector
- If len == 0.0 returns given vector (no unit vector possible)
- Else return calculated unit vector

Closes #23


src/check_line.c
```diff
@@ -67,6 +44,32 @@ bool	ft_isvalid_float_block(char **line, float min, float max)
...
+	checksum += ft_strtod(*line, &tmp);
+	if (!ft_isvalid_float(line, min, max, true))
+		return (false);
+	(*line)++;
+	checksum += ft_strtod(*line, &tmp);
+	if (!ft_isvalid_float(line, min, max, false))
+		return (false);
+	if (checksum == 0.0)
+		return (ft_perror_not_unit(begin));
+	else if (checksum > 1.0)
```
u413-284-si, line 71: A unit vector can have a checksum > 1.0. The definition of a vectors length is

sqr(x²+y²+z²)

The unit vector will have 1 = x² +  y ² + z² (as 1 to the power of 2 is still 1). So the checksum of the individual components to the power of 2 should be checked against > 1.0

Example: 
(0.905538, 0.3, 0.3) is perfectly a unit vector, but its simple checksum is obviously > 1

src/check_utils.c
```diff
@@ -0,0 +1,36 @@
+/* ************************************************************************** */
+/*                                                                            */
+/*                                                        :::      ::::::::   */
+/*   check_utils.c                                      :+:      :+:    :+:   */
```
u413-284-si: Naming: all utils files are names utils_something. 

In this case should be utils_check.c

src/vec3_linalgebra.c
```diff
@@ -19,7 +19,11 @@ float	ft_vec3_abs(t_vec3 v)
 
 t_vec3	ft_vec3_norm(t_vec3 v)
 {
-	return (ft_vec3_scale(v, 1.0 / ft_vec3_abs(v)));
+	const float	len = ft_vec3_abs(v);
+
+	if (len == 0.0)
```
u413-284-si: Is there any occurence of null vectors possible in the program runtime after implementing the check for it which aborts if one is given? If none can be introduced thereafter this check is superfluous

u413-284-si (changes requested):
The check for non-unit vectors is mathematically incorrect. The check for a  vectors length when normalising needs to be discussed regarding its necessity.

But yeah, it's good to catch these special cases in beforehand of the main processing.

src/check_line.c
gwolf-011235: True, sloppy from me!
Adapted: catch the return and add it squared to the checksum.

src/check_utils.c
gwolf-011235: Its a function which gets only called within the check stuff.

Can also rename the file check_move_line.c or something

src/vec3_linalgebra.c
gwolf-011235: I don't know. We also wouldn't immediately notice it, since its a float. This means div zero doesn't crash but would return infinity or NaN (standard float implementation has bits reserved for that)

In #23  we decided to add it. Can also remove it again.

u413-284-si: Agreed to remove that length check for performance gains

u413-284-si (approved):

PR #33, opened 2024-01-19
32 enable parsing of bonus scenes in base program and vice versa

### LINKED ISSUE #32: Enable parsing of bonus scenes in base program and vice versa
Currently base program will throw an error when parsing a bonus .rt file due to the additional information present. Vice versa the same applies.

u413-284-si:
Enables bonus program to read base files as well as bonus files. Base program only reads base files.

- added bonus functions for checker and parser
- differentiation achieved via presence of "#bonus" marker on top of scene files
- base checker and parser functions become common => accessible to bonus program
- boolean to choose between bonus/base functions for checker and parser

closes #32


gwolf-011235 (approved):
Looks good!

PR #34, opened 2024-01-22
27 bonus multithreading

### LINKED ISSUE #25: Key events are registered while printing
While the program prints to file (this is surprisingly slow!), other events should be blocked.

Right now the program is in the printing function but mlx still receives events. This means several prints can be started, which slows everything down. Or the scene could change which only happens after printing is done -> only then the scene will be re-rendered.

If it's not possible to clear pending events, print functionality will be deleted (or only bonus?)

### LINKED ISSUE #27: Bonus: Multithreading
The pixel colours are calculated independent of each other, which makes the calculations a good opportunity to implement multithreading.

We only use the CPU so the count of threads, which can be executed in parallel is limited.

To keep it simple we could fix the number of threads (e.g. 8) per Macro.

- [x] At every re-render of the scene create threads
- [x] Give every thread its package to work on. To even out the load every package could consist of all lines mod count = id. This means the image gets split in eight not evenly (first 100 lines), but consecutively (lines 1, 9, 17, ...).
- [x] Since every thread work on a different part of the image, no races should occur.
- [x] If thread creation fails program could exit or fall back to single threaded
- [x] printing in its own thread as to not impede showing stuff

gwolf-011235:
### Adds support for multi threading (Bonus)

- generic spin threads function which generates NUM_THREADS (miniRT_config) of threads
- threaded versions of raytracing, anti-alias, blend background, calc pix position
- can be toggled with 'J'

### Sampling (Bonus)

- the raytraced image is saved in a buffer (array of t_colour)
- before it is put to screen it is copied into the image
- for anti-aliasing the samples get added to the buffer and then divided by current sample size
- this continues until the sample_goal of SAMPLE_SIZE  is reached
- normal or anti_alias mode can be toggled with 'K'

### Detached thread for printing (Bonus)

- printing has been cut from base, since ignoring input while printing seems impossible
- instead a detached thread and a mutex is used.
- while printing a mutex protected bool is set, as well as a render option. This way we can ignore input and not continue to sample. When printing is finished, the bool is set to false and the option cleared
- Since printing has been excluded from base and in bonus we block input while printing Closes #25

### New menu page (Bonus)

- new menu page to display current running options (threads, anti-aliasing)

### Refactor

- instead of render_image > raytrace_image so we use the word raytrace somewhere :D
- norm everything (esp. render.h and render_bonus.h)
- various other small changes

Closes #27


inc/colour.h
```diff
@@ -65,16 +65,16 @@ t_colour	ft_hadamard_colour(t_colour c1, t_colour c2);
  * @brief Converts a t_colour struct to an int32_t.
  *
  * @param colour	Colour to convert.
- * @return int32_t	Converted colour.
+ * @return uint32_t	Converted colour.
```
u413-284-si, line 68: Why the change to unsigned?

inc/render_bit_field_bonus.h
```diff
@@ -0,0 +1,77 @@
...
+# define RENDER_BIT_FIELD_BONUS_H
+
+/* ====== LIBRARIES ====== */
+
+# include <stdint.h>
+# include <stdbool.h>
+
+/* ====== TYPEDEF ====== */
+
+typedef enum e_options
```
u413-284-si, line 23: So we could have a maximum of 32 options with the 32bits of uint32?

inc/threads_bonus.h
```diff
@@ -0,0 +1,84 @@
...
+	pthread_t	t_id;
+	void		*arg;
+}	t_thread_data;
+
+/* ====== FUNCTIONS ====== */
+
+/**
+ * @brief Creates NUM_THREADS threads and joins them.
+ *
+ * Routine can be any function which takes a void pointer as argument and
```
u413-284-si, line 38: Fancy mofo

src/parse_entity_sp_pl_cy.c
```diff
@@ -52,5 +52,8 @@ void	ft_parse_cylinder(char *line, t_hittable *cylinder, size_t id)
 	params->r = params->r / 2.0;
 	ft_parse_float(&line, &params->h);
 	ft_parse_colour_block(&line, &params->colour);
-	ft_cy_calc_caps(params);
+	params->cap1 = ft_vec3_add(params->centre,
```
u413-284-si, line 55: Why did you get rid of ft_cy_calc_caps again?

src/render_cleanup_bonus.c
```diff
@@ -0,0 +1,47 @@
...
+/*   By: gwolf <gwolf@student.42vienna.com>         +#+  +:+       +#+        */
+/*                                                +#+#+#+#+#+   +#+           */
+/*   Created: 2024/01/17 16:43:24 by gwolf             #+#    #+#             */
+/*   Updated: 2024/01/21 12:20:20 by gwolf            ###   ########.fr       */
+/*                                                                            */
+/* ************************************************************************** */
+
+#include "render_bonus.h"
+
+void	ft_free_mlx(t_mlx_ptrs *mlx_ptrs)
```
u413-284-si: The code here is 99% the same as in base part. According to our previous approach would make most functions common and then just differentiate between the ft_render_cleanup version with a conditional IS_BONUS

src/render_keyhook_options.c
```diff
@@ -64,12 +64,12 @@ void	ft_change_options(int key, t_render *render)
 {
 	if (key == XK_Escape)
 		mlx_loop_end(render->mlx_ptrs.mlx_ptr);
-	else if (key == XK_i)
-		ft_toggle_menu(&render->options, render->mlx_ptrs);
-	else if (key == XK_Control_L)
-		ft_change_mode(&render->options);
 	else if (key == XK_Shift_L)
 		ft_option_toggle(&render->options, O_SHOW_CTRL);
+	else if (key == XK_Control_L)
```
u413-284-si, line 69: Why the change in positioning?

src/render_output_ppm_bonus.c
```diff
@@ -34,47 +23,56 @@ static void	ft_write_ppm_header(int width, int height, int fd)
...
+		file_buffer[i * 3 + 2] = (*pixel & 0x0000FF);
+		pixel += 1;
+	}
+}
+
+t_err	ft_output_as_ppm(const t_img img)
 {
 	int		fd;
 	char	outfile[14];
+	char	*file_buffer;
```
u413-284-si, line 59: Why the extra buffer?

src/render_threaded_fts_bonus.c
```diff
@@ -0,0 +1,115 @@
...
+	ray.origin = render->cam.centre;
+	ray.d = 1.0;
+	pos.y = ((t_thread_data *)arg)->id;
+	while (pos.y < render->cam.image.y)
+	{
+		pos.x = -1;
+		while (++pos.x < render->cam.image.x)
+		{
+			colour = ft_shoot_aa_ray(pos, ray, render->scene, render->cam);
+			colour = ft_add_colour(colour,
```
u413-284-si: Could be shortened to one line 
render->[...] = ft_add_colour(colour, render->[...])

src/threads_bonus.c
```diff
@@ -0,0 +1,92 @@
...
+{
+	int	i;
+	int	ret;
+
+	i = -1;
+	while (++i < NUM_THREADS)
+	{
+		threads[i].id = i;
+		threads[i].arg = arg;
+		ret = pthread_create(&threads[i].t_id, NULL, routine, &threads[i]);
```
u413-284-si: Why not directly 
errno = pthread_create[...]

src/threads_bonus.c
```diff
@@ -0,0 +1,92 @@
...
+	}
+	return (SUCCESS);
+}
+
+void	ft_join_threads(t_thread_data threads[NUM_THREADS], bool join_all)
+{
+	int	i;
+
+	i = -1;
+	if (join_all)
```
u413-284-si: Is that necessary? 

Just the second loop would cover all cases, no?

src/threads_bonus.c
```diff
@@ -0,0 +1,92 @@
...
+	thread.arg = arg;
+	errno = pthread_create(&thread.t_id, NULL, routine, &thread);
+	if (errno)
+	{
+		ft_perror("pthread_create() failed", errno);
+		return (ERROR);
+	}
+	errno = pthread_detach(thread.t_id);
+	if (errno)
+		ft_perror("pthread_detach() failed", errno);
```
u413-284-si, line 78: why no ERROR return here?

u413-284-si (changes requested):
Great work, especially implementing the differentiation with the different cases when threading, anti-aliasing and all its combinations. Also nicely done with the void pointers/function pointers.

Just couple of minor questions and/or change requests.

inc/colour.h
gwolf-011235: Seems to be more inline with standard way of doing things. Has no effect on program since it's just representation of the underlying data > we don't care about the actual number but about the r,g,b channels

inc/render_bit_field_bonus.h
gwolf-011235: Yes - one bit one option

inc/threads_bonus.h
gwolf-011235: :man_dancing:

src/parse_entity_sp_pl_cy.c
gwolf-011235: Refactor: It was one function in a separate file. Since it was just two lines, i put them there directly

src/render_output_ppm_bonus.c
gwolf-011235: Speeds up printing immensely.
Instead of writing each pixel value line by line to the file, we write the r,g,b values at first to this (local) buffer. After it has been written, the buffer can be written to file in one go.

src/threads_bonus.c
gwolf-011235: True! Changed

src/render_threaded_fts_bonus.c
gwolf-011235: `render->[...] = ft_add_colour(`
This line is already too long. Would need to rename variable sample_buffer to only buffer or something

src/threads_bonus.c
gwolf-011235: It's like a close() error: we can't really do anything about it. 

It "shouldn't" happen as pthread_detach() only errors if the thread could not be found or is not joinable. Both conditions should be true since we successfully created a standard (= joinable) thread.

src/render_keyhook_options.c
gwolf-011235: To group the keys better together: Shift and Control are right next to each other

src/parse_entity_sp_pl_cy.c
u413-284-si: Back to base

src/render_output_ppm_bonus.c
u413-284-si: So the writing process is itself time consuming

src/render_threaded_fts_bonus.c
u413-284-si: Yeah, I meant you dont need to assign it to colour and then assign colour to render->[...]

src/render_cleanup_bonus.c
gwolf-011235: The functions are mostly syntactic sugar, I got get rid of them entirely. Only ft_free_mlx() and render_cleanup() are needed. 

In render_cleanup both are, where render_cleanup has an ifndef guard.
In render_cleanup_bonus only render_cleanup is there.

src/render_output_ppm_bonus.c
gwolf-011235: Yes!
Writing to file is way slower than malloc - "write" to buffer - write to file.
Writing to file needs the syscall write for every line.
Buffer only needs one write() call as "writing" to buffer is not exactly writing but just setting the char array.

src/render_threaded_fts_bonus.c
gwolf-011235: `render->sample_buffer[pos.y * render->cam.image.x + pos.x] = ft_add_colour(`
is too long.

Pushing the function to the next line also does not work, because then the last line is too long
```
render->sample_buffer[pos.y * render->cam.image.x + pos.x] =
   ft_add_colour(colour,
         render->sample_buffer[pos.y * render->cam.image.x + pos.x]);;
```

Or am I missing something?

src/threads_bonus.c
gwolf-011235: That is correct!
I set the ID to -1 after creating this function, didn't update accordingly

src/render_threaded_fts_bonus.c
u413-284-si: You could line break again after the asteriks *

src/threads_bonus.c
u413-284-si: So we don't need the boolean join_all

src/threads_bonus.c
u413-284-si: k

src/threads_bonus.c
gwolf-011235: Yeah, last message was unclear: it deleted the extra loop and the boolean

src/render_threaded_fts_bonus.c
gwolf-011235: Which would give us 4 lines instead of 3..

Performance wise it makes no difference, as the compiler will optimize away (even if not, would be in the space of microseconds).
Readability wise it gets more confusing as one assignment spans over 4 lines.

If you really don't like it I can rename the sample_buffer to buffer but I don't like the 4 line assignment.

u413-284-si: naaa, I meant like
```
render->sample_buffer[pos.y * render->cam.image.x + pos.x] =
   ft_add_colour(colour, render->sample_buffer[pos.y * \
   render->cam.image.x + pos.x]);
```
It's not for performance, but it's a hiccup to see the consecutive assignment of colour = sth.1 and the next line sth.2 = colour instead of sth2 = sth1

gwolf-011235:
```
colour = ft_shoot_aa_ray(pos, ray, render->scene, render->cam);
render->sample_buffer[pos.y * render->cam.image.x + pos.x]
	= ft_add_colour(colour, render->sample_buffer[pos.y * \
		render->cam.image.x + pos.x]);
```
v
```
colour = ft_shoot_aa_ray(pos, ray, render->scene, render->cam);
colour = ft_add_colour(colour,
		render->sample_buffer[pos.y * render->cam.image.x + pos.x]);
render->sample_buffer[pos.y * render->cam.image.x + pos.x] = colour;
```

I don't get it. Changed.

u413-284-si (approved):

PR #39, opened 2024-01-23
38 remove cam from specular lighting calculation

### LINKED ISSUE #38: Remove cam from specular lighting calculation
t_cam is used to calculate the viewers direction. The same calculation can be effected by using a parameter already present in the hitrecord making it obsolete to pass t_cam in the scene light functions.

u413-284-si:
Implements two performance optimisations:
1)

Replace cam->centre with rec.ray.origin to calculate viewers direction in specular lighting.
Allows omition of t_cam in the whole colour calculation starting from ft_ray_colour downwards. Saves copying t_cam struct at each function call.

2)
Also replaced all occurences of pow(x,2) with x * x. Should allow for faster processing.

closes #38


gwolf-011235 (approved):
Looks good!

After merge of multithread we can get rid of ray_bonus, since they are the same (i moved the sample generation to another bonus file).

PR #40, opened 2024-01-24
37 bonus cone manipulation

### LINKED ISSUE #37: Bonus: Cone manipulation
As with the other hittables: cone be manipulated by keypresses if in mode scene

gwolf-011235:
- Adds manipulation function for cone following scheme of other manip fts
- Fixes bug I introduces by deleting ray_bonus.c - file restored.
- Fix camera rotation speed: after change to quaternion for rotation was too fast. Also removed doubled code via ifndef


src/render_keyhook_camera.c
```diff
@@ -15,13 +15,13 @@
 void	ft_keyhook_rot_cam(int key, t_vec3 *direction, float inc)
 {
 	if (key == XK_Left)
-		*direction = ft_vec3_rotate_y(*direction, inc);
+		*direction = ft_vec3_rotate_y(*direction, inc * 0.1);
```
u413-284-si: Can we not decrease inc at its source by a decimal point?

u413-284-si (changes requested):
Very good! Are we now fast enough to see keypress changes with bonus scenes more or less fluently?

Just one question regarding the movement increment.

src/render_keyhook_camera.c
gwolf-011235: Then the increment would be too low for movement, since it is also used there. 
For object rotation it's not that noticable, but camera can be way off especially in slower (base) mode.

But I changed the rotation to use degrees instead of radians (by converting) which has a similiar effect. Object rotation is now also slower but acceptable in my opinion

gwolf-011235:
In bonus I run at about 4-5 with AA and about 25 without. So yes it's more fluid

u413-284-si (approved):

PR #42, opened 2024-01-25
41 bad access if empty scene

### LINKED ISSUE #41: Bad access if empty scene
If there are no objects in the scene, manip functions should not try to access hittable array. Also menu text should not try to access not existing object.

Another thing is that mlx_set_font() seems to leak. Although other font looks nicer, function should be eliminated to avoid confusion

gwolf-011235:
If no object array is not malloced and active hittable is set to -1
If select key is pressed, check for -1 to leave early
If manip key is pressed, and mode scene: check for -1 to leave early
In menu check for -1: if so write "No objects in scene" instead of stats.

Removed mlx_set_font and associated macros


u413-284-si (approved):
Looking guut! No further comments

u413-284-si:
Has some conflicts xD

PR #43, opened 2024-01-26
26 bonus parse window size

### LINKED ISSUE #26: Bonus: parse window size
The window size is fixed with macro WIN_SIZE_X and WIN_SIZE_Y. 
We could implement a new identifier (e.g. 'S' for size), where the line consists of two int values: X and Y.
```
S 800 600
```

Should be relatively easy to implement. 

- [ ] Accept new identifier 'S'
- [ ] New int check function (currently only accepts 0 to 255 for rgb)
- [ ] Allowed values should take into account size of menu > it should not be to small (or disable menu if to small)

gwolf-011235:
### Implements parsing for Window size
- only works in "#bonus" files
- win_size line is identified by 'S'
- must contain two integers in range WIN_SIZE_X/Y and MAX_WIN_SIZE_X/Y. Must be separated by comma
- checks if current screen is bigger than accepted win_size with MLX  function. If so sets it to the received value.
- adapted parsing functions to also receive win_size. In base this gets set to WIN_SIZE_X/Y. In bonus if no line received also gets set to the same value.

Closes #26


u413-284-si (approved):
Splendid!

PR #45, opened 2024-01-28
44 refactor double header code

### LINKED ISSUE #44: Refactor: double header, code
Header files often have the same defintions for base and bonus. Also some source files contain doubled code.

gwolf-011235:
### Reduce duplication

- Many of the bonus headers were adapted: include the base header and build on that. This means that only one prototype exists of a function and not two parallel ones.
- Code which was present in more than one file has been deleted to reduce possibilities for errors. 
- Render struct had to be extracted into it's own file for this to work.
- Random has also been extracted since its the only bonus stuff left in utils.

### Clearer Module for manipulation

- Manipulation functions were extracted from big header render.h and put into their own header manip.h. 
- Belonging files got prefix "manip_" to distinguish them

### Doxygen update

- To only include files which are used (base or bonus) all base files got postfix _base.c
- Doxygen filters these out
- Right now only bonus documentation is generated since a way to put both base and bonus on github pages seems complicated
- Small: Doxygen file updated to not sort members of struct.


u413-284-si (approved):
No inconsistencies discovered. Refactor looks solid!

PR #48, opened 2024-07-03
fix: set ptr to NULL to avoid double free

### LINKED ISSUE #47: Crash if mlx can't be instantiated
Current logic crashes if mlx init fails, because of double free. Pointers need to be set to NULL.

gwolf-011235:
Fixes crash detected by funcheck

Closes #47


u413-284-si (approved):
Seems correct. So you managed to crash it by failing mlx_init()?

